PR #1, opened 2017-12-06
grpc connection management no longer yields same errors from dial

jhump:
When the test was written, `grpc.Dial` returned underlying connect error on failure/timeout. Now (sadly) it only returns a context cancelled/timeout error.

Also, go "tip" version issues compile errors for format pattern mistakes and caught a problem with an error message that was using `"%d"` for a string value.

A considerable downside of this is that all connection and TLS exceptions simply get reported as timeouts 😦. I'll also be filing a bug in the GRPC project about this sudden inability to access the cause of connection errors.

### DIFF
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -854,9 +854,6 @@ func BlockingDial(ctx context.Context, address string, creds credentials.Transpo
 	// grpc.Dial doesn't provide any information on permanent connection errors (like
 	// TLS handshake failures). So in order to provide good error messages, we need a
 	// custom dialer that can provide that info. That means we manage the TLS handshake.
-	var wg sync.WaitGroup
-	wg.Add(1)
-	var once sync.Once
 	result := make(chan interface{}, 1)
 
 	writeResult := func(res interface{}) {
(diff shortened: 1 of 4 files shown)


dragonsinth:
I don't really know all the context, but it looks like if you shim the Dialer with your own, maybe you could stash the real error for testing/reporting?

jhump:
@dragonsinth, I changed this to do what you suggest. We provide our own dialer so that we can report better error messages faster -- both in tests and in the command-line tool (so users don't have to wait the full connect timeout and get nothing better than "deadline exceeded" error).

grpcurl.go
```diff
@@ -844,3 +846,75 @@ func ServerTransportCredentials(cacertFile, serverCertFile, serverKeyFile string
...
+// BlockingDial is a helper method to dial the given address, using optional TLS credentials,
+// and blocking until the returned connection is ready. If the given credentials are nil, the
+// connection will be insecure (plain-text).
+func BlockingDial(ctx context.Context, address string, creds credentials.TransportCredentials, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
+	// grpc.Dial doesn't provide any information on permanent connection errors (like
+	// TLS handshake failures). So in order to provide good error messages, we need a
+	// custom dialer that can provide that info. That means we manage the TLS handshake.
+	var wg sync.WaitGroup
+	wg.Add(1)
+	var once sync.Once
```
dragonsinth: I don't think these 3 are actually used, you don't `wg.Wait()` anywhere

jhump: Doh! Yep, I forgot to remove that shit. They were part of an earlier formulation, before I realized to that I need to use a channel.

PR #3, opened 2018-02-05
make jsonpb marshaler emit defaults

srenatus:
This protobuf default is really quite confusing:

Example call before:

```
$ echo '{"subjects": ["teams:admins2"], "resource": "authz:nodes:foo:run_list", "action": "nodes:read" }' | grpcurl -v -d @ localhost:9094 authz.Authorization/IsAuthorized
Resolved method descriptor:
{
  "name": "IsAuthorized",
  "inputType": ".authz.IsAuthorizedReq",
  "outputType": ".authz.IsAuthorizedResp",
  "options": {
  }
}
Request metadata to send:
(empty)
Response headers received:
content-type: application/grpc
Response contents:
{

}
```

This happened because defaults are omitted, but for a boolean response, the `false` case actually is quite interesting! 😄 

So, with this, we'd have grpcurl emit defaults, and the crucial section of the output becomes:
```
Response contents:
{
  "authorized": false
}
```

What do you think?

I could be persuaded to turn this into a flag; but that seems to require some effort to thread the config value through, so I wanted to propose the simplest thing first. 😉 

Thanks for grpcurl! It's helped our team quite a bit. 👍

### DIFF
--- a/grpcurl_test.go
+++ b/grpcurl_test.go
@@ -261,6 +261,20 @@ const (
     "body": "SXQncyBCdXNpbmVzcyBUaW1l"
   }
 }`
+	fullResponse1 = `{
+  "payload": {
+    "type": "COMPRESSABLE",
+    "body": "SXQncyBCdXNpbmVzcyBUaW1l"
+  },
+  "username": "",
+  "oauthScope": ""
+}`
+	response1 = `{
+  "payload": {
+    "type": "COMPRESSABLE",
+    "body": "SXQncyBCdXNpbmVzcyBUaW1l"
+  }
+}`
 	payload2 = `{
   "payload": {
     "type": "RANDOM",
@@ -292,7 +306,7 @@ func doTestUnary(t *testing.T, cc *grpc.ClientConn, source DescriptorSource) {
 	}
 
 	if h.check(t, "grpc.testing.TestService.UnaryCall", codes.OK, 1, 1) {
-		if h.respMessages[0] != payload1 {
+		if h.respMessages[0] != fullResponse1 {
 			t.Errorf("unexpected response from RPC: expecting %s; got %s", payload1, h.respMessages[0])
 		}
 	}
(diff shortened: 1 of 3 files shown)


srenatus:
The tests for 1.6 and 1.7 seem to be a bit flaky. I wonder if it's worth keeping them around anyways. 🤔

grpcurl_test.go
```diff
@@ -260,6 +260,20 @@ const (
...
+}`
+	fullResponse1 = `{
+  "payload": {
+    "type": "COMPRESSABLE",
+    "body": "SXQncyBCdXNpbmVzcyBUaW1l"
+  },
+  "username": "",
+  "oauthScope": ""
+}`
+	response1 = `{
```
jhump: Why retain this constant? I see it used for half-duplex stream test. Why does that test behave differently than the unary test?

jhump:
> The tests for 1.6 and 1.7 seem to be a bit flaky.

gRPC still supports 1.6, and there's nothing in the code that won't work with 1.6. I suspect the reason gRPC (and protobuf and other libraries) still support that old version might be the Google App Engine standard runtime for Go, which only recently offered people a path to upgrade from Go 1.6 (straight to 1.8).

I haven't seen 1.7 act flaky, but I have definitely seen that flaky failure in 1.6 (`TestPlainText` in `tls_settings_test.go`). We'll fix it at some point, possibly even by dropping 1.6 from the test matrix.

grpcurl_test.go
srenatus: > Why does that test behave differently than the unary test?

Heh, to be quite honest, I have no idea, and it was surprising me as well. Can you tell?

srenatus: Anyways, this will probably go away with the re-work to make this a flag.

srenatus:
I've turned it into a flag. To have that be effective, the json marshalling now happens in `cmd/grpcurl`. I think that's probably ok. What do you think?
(I'll of course squash this when we've reached a conclusion.)

PR #5, opened 2018-02-17
Add Makefile

peter-edge:
This is a semi-standard Makefile I use for a lot of my projects, I find it helps a lot to have, especially to show new developers what development commands to execute. Note `golint, errcheck, staticcheck` don't pass yet, but I aim to do that shortly (I can remove from this PR if you want). Let me know if this fits your style, this is just a suggestion.


jhump:
@peter-edge, this looks great. But it's not passing.

Strangely, just adding a `Makefile` to the repo causes Travis to skip its default install step: https://docs.travis-ci.com/user/languages/go/

So I think you just need another make step named "install" that will run it: `go get -t -v ./...`

peter-edge:
I actually copied that Makefile from a project I have vendor checked in to, I added `make deps` and `make updatedeps` which are two of my standard targets for repositories without vendor directories, and added `deps` to the travis build.

PR #8, opened 2018-02-19
attempt to fix flaky test

jhump:
The tests are usually failures to setup the connecting within 1 second. It happens much more often in Go 1.6 than others, but still happens in others (so just dropping support for 1.6, which is certainly a consideration, wouldn't truly fix the flakiness).

Sometimes the error is during the RPC call, due to transient connection issue. So turning off fail-fast should address that. Another error I've seen a couple of times is in `TestServerDoesNotSupportReflection`, due to getting an `io.EOF` error instead of the expected `ErrReflectionNotSupported`. That is a bug in grpcreflect (dependency), and it should be fixed by this patch: https://github.com/jhump/protoreflect/pull/104


jhump:
Well damn. Test is still flaky. (Got a flaky failure when this merged to master 😞.)

PR #13, opened 2018-02-20
fix golint

chemidy:
just fix some golint warning

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ matrix:
     - go: 1.7
     - go: 1.8
     - go: 1.9
+    - go: "1.10"
     - go: tip
 
 script:


grpcurl.go
```diff
@@ -345,9 +344,8 @@ func InvokeRpc(ctx context.Context, source DescriptorSource, cc *grpc.ClientConn
 		return invokeClientStream(ctx, stub, mtd, handler, requestData, req)
 	} else if mtd.IsServerStreaming() {
 		return invokeServerStream(ctx, stub, mtd, handler, requestData, req)
-	} else {
-		return invokeUnary(ctx, stub, mtd, handler, requestData, req)
 	}
+	return invokeUnary(ctx, stub, mtd, handler, requestData, req)
```
jhump: This being at a different indentation level from the others looks funny to me. I bet we can change that and still keep the linter happy by changing this to a `switch` where this is the default case.

cmd/grpcurl/grpcurl.go
```diff
@@ -244,7 +244,9 @@ func main() {
 			refClient = nil
 		}
 		if cc != nil {
-			cc.Close()
+			if err := cc.Close(); err != nil {
+				fail(err, "Failed to close grpc Client")
```
jhump: This doesn't seem useful. At this point, the operation is complete, so errors here can be suppressed.

If the linter is complaining that you are ignoring the return value, perhaps it would allow:

```
_ = cc.Close()
```

grpcurl.go
```diff
@@ -276,7 +276,7 @@ type InvocationEventHandler interface {
 // the supplier has no more messages, it should return nil, io.EOF.
 type RequestMessageSupplier func() ([]byte, error)
 
-// InvokeRpc uses te given GRPC connection to invoke the given method. The given descriptor source
+// InvokeRPC uses te given GRPC connection to invoke the given method. The given descriptor source
```
jhump: Please don't change exported APIs in breaking ways.

testing/test_server.go
```diff
@@ -15,9 +15,10 @@ import (
 	"github.com/fullstorydev/grpcurl"
 )
 
+// TestServer is the testserver instance
```
jhump: Not a very helpful comment. If you're going to add a doc comment, please say something a little more useful.

testing/test_server.go
```diff
@@ -15,9 +15,10 @@ import (
...
 type TestServer struct{}
 
-// One empty request followed by one empty response.
+// EmptyCall is One empty request followed by one empty response.
```
jhump: Bad capitalization: s/is One/is one/
(Here and below/throughout.)

testing/test_server.go
```diff
@@ -181,10 +182,10 @@ func (TestServer) HalfDuplexCall(str grpc_testing.TestService_HalfDuplexCallServ
 			return str.Context().Err()
 		}
 		if req, err := str.Recv(); err != nil {
-			if err == io.EOF {
-				break
+			if err != io.EOF {
```
jhump: Weird. The linter complained about the order of breaks vs. returns?

testing/test_server.go
```diff
@@ -203,6 +204,7 @@ func (TestServer) HalfDuplexCall(str grpc_testing.TestService_HalfDuplexCallServ
 	return nil
 }
 
+// Metadata const
```
jhump: please remove

.travis.yml
```diff
@@ -7,8 +7,9 @@ matrix:
     - go: 1.7
     - go: 1.8
     - go: 1.9
+    - go: "1.10"
```
jhump: Mind making all of them strings, for consistency?

Alternatively, you might change all of them to use ".x" suffix (so yaml parser will interpret all of them as strings). E.g. 1.7.x, 1.8.x, etc. Travis understands this suffix (and it will maintain same semantics as current: latest point release for request major.minor version).

Issue #14, opened 2018-02-26
Include headers on discovery calls


FreGalle:
Headers are currently not sent along with service/method discovery calls. This prevents grpcurl from being used with any server that for example performs custom header-based authentication.

Any particular reasoning behind the decision not to send headers with the discovery calls? I think it would be sensible for any server performing (custom) authentication/authorization before executing a service call to extend the auth to the discovery of its services.

jhump:
That seems reasonable. I think it should be behind a command-line option -- either a flag that indicates that metadata should accompany service reflection requests or options to specify a separate set of headers just for the reflection calls. I think the former would be the simplest approach.

If you want to take a crack at a PR, I'll take a look.

FreGalle:
Sure, I'll have a look and probably open a PR later this week.

PR #15, opened 2018-02-27
Allow reflection calls to carry headers

FreGalle:


cmd/grpcurl/grpcurl.go
```diff
@@ -54,6 +54,9 @@ var (
     	read from stdin. For calls that accept a stream of requests, the
     	contents should include all such request messages concatenated together
     	(optionally separated by whitespace).`)
+	reflectWithHeaders = flag.Bool("reflect-with-headers", false,
+		`Include headers provided with -H when performing reflection calls
+		against a remote server.`)
```
jhump: Do you mind double-checking how this looks when you do `grpcurl -h`? I remember the formatting of flag help in the `flag` package is not great, so I had to manually format the other remarks in here so they rendered properly. For example, this second line may need lower level of indent, like the `data` parameter above.

cmd/grpcurl/grpcurl.go
```diff
@@ -155,9 +158,6 @@ func main() {
 		if *data != "" {
 			fail(nil, "The -d argument is not used with 'list' or 'describe' verb.")
 		}
-		if len(addlHeaders) > 0 {
```
jhump: This seems confusing. If we allow `-H` flag, even if they do not specify `-reflect-with-headers`, then the user will likely be confused that the `list` and `describe` operation are *not* actually using the headers.

I can think of a few possible alternatives:
1. Restore this check if `!*reflectWithHeaders`.
2. Allow `-H` and simply don't require `-reflect-with-headers` with the `list` and `describe` verbs. I think you can achieve that by just setting `*reflectWithHeaders = true` here.
3. Use a separate flag from `-H` for the reflect headers. Then we can still disallow `-H` here and use headers defined by the other flag when creating the reflection client.

Of these, I think option 1 is the crummiest. Even though option 3 differs from what I said on the bug (#14), it suddenly seems the best just to make the behavior more coherent across all of the "verbs" that grpcurl supports.

WDYT?

FreGalle: I agree that the behavior of `grpcurl -H ... describe ...` is rather confusing now when you expect headers should only have to be provided once.

I thought about option 1, but decided against it as it would generate as much confusion as it would prevent. Option 3 I'm not a big fan of: having to provide the same set of headers twice when you want to send them for all calls is neither intuitive nor a great user experience.

If I understand you correctly, option 2 is the equivalent of not even bothering with an extra flag, but always using headers for invocation and reflection calls. To me, this would be the most intuitive behaviour right out of the box. Setting headers (as a user) means you expect them to be used.

cmd/grpcurl/grpcurl.go
FreGalle: That's correct. I'll commit the fixed indentation as soon as we settle on a solution for the other issue.

FreGalle: FYI, in Go 1.10 there's [an improvement](https://golang.org/doc/go1.10#flag) that provides better alignment of flag usage strings when including newlines. The current output on 1.10 is therefore funky again for all flags.

cmd/grpcurl/grpcurl.go
jhump: Yes, option 2 is the equivalent of not bothering with an extra flag, but *only* for `list` and `describe` actions (e.g. when there are *only* reflective calls being done).

I suppose there is a fourth option, which provides the same coherence of option 3 above, but without requiring you to redundantly specify the same headers for both reflective calls and the main RPC call:

1. `-H` means headers for all requests, including reflection.
2. Add a new `-request-header` flag that is used *only* for the main RPC invocation. It is an error to specify `-request-header` with the `list` and `describe` verbs.
3. Add a new `-reflect-header` flag that is used *only* for reflective calls. It is an error to specify `-reflect-header` if using protoset files instead of reflection.

The effective headers for a request will end up merging these flags. So RPC actions will use `-H` *plus* `-request-header` values; reflection calls will use `-H` *plus* `-reflect-header` values.

What do you think?

FreGalle: First off: if at all possible, a user shouldn't have to think about whether an invocation of `grpcurl` will perform any reflection calls or not (in my opinion).

In light of that, I'm not a big fan of option 2 (thanks for the nuance). Flags that are only meaningful for certain subcommands but not for others are not well-designed (unless the flag is only mentioned in the documentation of the relevant subcommands).

I do like your new take on option 3. It is definitely the most expressive as it allows a user to specify precisely which headers he wants to include in which types of requests. The simplest case remains to only specify `-H` and for most users, this should just work.

jhump: Okay. I guess it just feels weird to send unnecessary headers to the reflect calls. Other than general auth to get into a service, I would expect a lot of other metadata to be service-specific. It just feels like a potential security risk to be sending metadata (which could include things like tokens unused by the reflective service, etc) for requests that don't need/want it. Not that the reflection service will do anything with it, and it is the same target host:port as the real requests.. it just seems not-great.

For now, let's just remove this new flag and always send the `-H` headers for both invocations and reflective calls. And, if you don't mind, add a TODO for add'l flags to better control what we send where.

FreGalle: Sending the headers to both reflection and invocation services is a slight security risk but one I feel is worth the trade-off in terms of improved usability.

PR #17, opened 2018-03-08
add more control over request metadata; add message templates

jhump:


cmd/grpcurl/grpcurl.go
```diff
@@ -157,6 +171,9 @@ func main() {
 		if *data != "" {
 			fail(nil, "The -d argument is not used with 'list' or 'describe' verb.")
 		}
+		if len(rpcHeaders) > 0 {
+			fail(nil, "The -rpc-header argument is not used with 'list' or 'describe' verb.")
+		}
```
dragonsinth: Maybe just warn?  I found when I was using grpcurl repeatedly, I wanted to keep mostly the same command line intact and just change around the bare minimum.  For example:

```
./bin/grpcurl -v -org=foo -d '{ "id": "daniel.falko@example.com" }' 127.0.0.1:9240 list`
./bin/grpcurl -v -org=foo -d '{ "id": "daniel.falko@example.com" }' 127.0.0.1:9240 list fullstory.users.Users`
./bin/grpcurl -v -org=foo -d '{ "id": "daniel.falko@example.com" }' 127.0.0.1:9240 describe fullstory.users.Users.GetUser
./bin/grpcurl -v -org=foo -d '{ "id": "daniel.falko@example.com" }' 127.0.0.1:9240 fullstory.users.Users.GetUser
```

Only the last call actually needed `-org` and `-d`, but it would have been kind of annoying if the tool had failed on me for providing them.  I kinda feel that way about rpcHeaders / reflHeaders here.

jhump: Okay. Changed a few places to log warnings about the args being ignored instead of aborting

PR #23, opened 2018-03-23
enable all linters in CI; fix lint issues

jhump:
Most of the lint complaints were either the form (1) `if` block has a return statement so remove `else` block and outdent it (2) or function returns error value that is being ignored. For the latter, many of the cases were appropriate to ignore the error, so I just added `_ = ` to "consume" the error and get rid of the lint failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,64 +1,65 @@
-SRCS := $(shell find . -name '*.go')
-PKGS := $(shell go list ./...)
-
-.PHONY: all
-all: deps lint test
+# TODO: run golint and errcheck, but only to catch *new* violations and
+# decide whether to change code or not (e.g. we need to be able to whitelist
+# violations already in the code). They can be useful to catch errors, but
+# they are just too noisy to be a requirement for a CI -- we don't even *want*
+# to fix some of the things they consider to be violations.
+.PHONY: ci
+ci: deps checkgofmt vet staticcheck unused ineffassign predeclared test
 
 .PHONY: deps
 deps:
-	go get -d -v -t $(PKGS)
+	go get -d -v -t ./...
 
 .PHONY: updatedeps
 updatedeps:
-	go get -d -v -t -u -f $(PKGS)
+	go get -d -v -t -u -f ./...
 
 .PHONY: install
 install:
-	go install $(PKGS)
-
-.PHONY: golint
-golint:
-	@go get github.com/golang/lint/golint
-	for file in $(SRCS); do \
-		golint $${file}; \
-		if [ -n "$$(golint $${file})" ]; then \
-			exit 1; \
-		fi; \
-	done
+	go install ./...
 
 .PHONY: checkgofmt
 checkgofmt:
-	gofmt -s -l $(SRCS)
-	if [ -n "$$(gofmt -s -l $(SRCS))" ]; then \
+	gofmt -s -l .
+	@if [ -n "$$(gofmt -s -l .)" ]; then \
 		exit 1; \
 	fi
 
 .PHONY: vet
 vet:
-	go vet $(PKGS)
-
-.PHONY:
-errcheck:
-	@go get github.com/kisielk/errcheck
-	errcheck $(PKGS)
+	go vet ./...
 
 .PHONY: staticcheck
 staticcheck:
 	@go get honnef.co/go/tools/cmd/staticcheck
-	staticcheck $(PKGS)
+	staticcheck ./...
 
 .PHONY: unused
 unused:
 	@go get honnef.co/go/tools/cmd/unused
-	unused $(PKGS)
+	unused ./...
 
-.PHONY: lint
-lint: golint checkgofmt vet errcheck staticcheck unused
+.PHONY: ineffassign
+ineffassign:
+	@go get github.com/gordonklaus/ineffassign
+	ineffassign .
+
+.PHONY: predeclared
+	@go get github.com/nishanths/predeclared
+	predeclared .
+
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,64 +1,56 @@
-SRCS := $(shell find . -name '*.go')
-PKGS := $(shell go list ./...)
```
jhump: These seemed a little odd since none of the tools actually needed them. This came from a contribution/pull request. Maybe their intent was to speed things up, to only scan the file system once ? Seems totally unnecessary, and removing bash loop from the calls below makes me like the lay of the land a little more.

Makefile
```diff
@@ -1,64 +1,56 @@
...
 .PHONY: golint
 golint:
 	@go get github.com/golang/lint/golint
-	for file in $(SRCS); do \
-		golint $${file}; \
-		if [ -n "$$(golint $${file})" ]; then \
-			exit 1; \
-		fi; \
-	done
+	golint -min_confidence 0.9 -set_exit_status ./...
```
jhump: No need for double-running `go lint` if we use `-set_exit_status`. (Unfortunately, `gofmt` has no such flag.)

Also, the default confidence level of 0.8 caused it to complain about an exported method name. I don't like the idea of changing names in exported API just to satisfy a linter, so I up'ed this to make the warning go-away. (Unfortunately, it doesn't seem like this tool has other knobs for making exceptions and/or white-listing violations)

testing/test_server.go
```diff
@@ -204,10 +207,26 @@ func (TestServer) HalfDuplexCall(str grpc_testing.TestService_HalfDuplexCallServ
 }
 
 const (
-	MetadataReplyHeaders  = "reply-with-headers"
+	// MetadataReplyHeaders is a request header that contains values that will
+	// be echoed back to the client as response headers. The format of the value
+	// is "key: val". To have the server reply with more than one response
+	// header, supply multiple values in request metadata.
+	MetadataReplyHeaders = "reply-with-headers"
```
jhump: The linter complained about exported symbols that had no Go doc, so I fleshed these out.

cmd/grpcurl/grpcurl.go
```diff
@@ -269,7 +269,7 @@ func main() {
 			refClient = nil
 		}
 		if cc != nil {
-			cc.Close()
+			_ = cc.Close()
```
dragonsinth: yick

cmd/grpcurl/grpcurl.go
```diff
@@ -235,7 +235,7 @@ func main() {
 				fail(err, "Failed to configure transport credentials")
 			}
 			if *serverName != "" {
-				creds.OverrideServerName(*serverName)
+				_ = creds.OverrideServerName(*serverName)
```
dragonsinth: warn?

grpcurl.go
```diff
@@ -225,16 +225,16 @@ func ListMethods(source DescriptorSource, serviceName string) ([]string, error)
...
-		}
-		sort.Strings(methods)
-		return methods, nil
 	}
+	methods := make([]string, 0, len(sd.GetMethods()))
+	for _, method := range sd.GetMethods() {
+		methods = append(methods, method.GetName())
+	}
+	sort.Strings(methods)
+	return methods, nil
```
dragonsinth: In general, I kind of dislike a linter forcing you to do this; if you really don't need `sd` beyond the scope of the else statement.

grpcurl.go
```diff
@@ -345,9 +344,8 @@ func InvokeRpc(ctx context.Context, source DescriptorSource, cc *grpc.ClientConn
 		return invokeClientStream(ctx, stub, mtd, handler, requestData, req)
 	} else if mtd.IsServerStreaming() {
 		return invokeServerStream(ctx, stub, mtd, handler, requestData, req)
-	} else {
-		return invokeUnary(ctx, stub, mtd, handler, requestData, req)
 	}
+	return invokeUnary(ctx, stub, mtd, handler, requestData, req)
```
dragonsinth: I just find this so useless and less readable :(

grpcurl_test.go
```diff
@@ -62,7 +66,9 @@ func TestMain(m *testing.M) {
 		grpc.WithInsecure(), grpc.WithBlock()); err != nil {
 		panic(err)
 	}
-	defer ccReflect.Close()
```
dragonsinth: ok... this is where I draw the line, this is horrible.  Not better

grpcurl_test.go
```diff
@@ -87,7 +97,9 @@ func TestMain(m *testing.M) {
 		grpc.WithInsecure(), grpc.WithBlock()); err != nil {
 		panic(err)
 	}
-	defer ccProtoset.Close()
+	defer func() {
+		_ = ccProtoset.Close()
+	}()
```
dragonsinth: lk;ajdsflk;jsdafl;ksdajf

cmd/grpcurl/grpcurl.go
jhump: Yeah, I know.

cmd/grpcurl/grpcurl.go
jhump: Doh! I was mechanically ignoring errors, but this is a real violation that needs to be fixed!

jhump:
@dragonsinth, I removed `golint` and `errcheck` from the tools run in CI. They are what produced the more egregious constructs that were in this PR. I reverted them and just kept the ones that I feel are improvements (or at worst, neutral).

PTAL

PR #35, opened 2018-05-29
fix race condition in bidi tests

jhump:
Strangely, this race condition was tickled by the change I made to allow use of proto sources (not just compiled protoset decsriptor files). But, after evaluating the problem, it was technically racy for all of the kinds of stubs (protosets, proto sources, and those that used server reflection).

It is possible (especially I'd never seen the issue before) that the problem was actually caused by a recent change in the gRPC library.

The problem was that a goroutine was cancelling the stream context after an error was observed on the stream. There are several paths where gRPC stream operations can then use that context error instead of the original stream error. So the test code was expecting a "permission denied", but would occasionally instead see "cancelled".

It turns out, cancelling the context is wholly unnecessary as the stream is already aborted whenever a send error is observed. So just removing the call to `cancel()` should fix the raciness, and all of the tests still pass without this call.

### DIFF
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -386,7 +386,7 @@ func InvokeRpc(ctx context.Context, source DescriptorSource, cc *grpc.ClientConn
 	defer cancel()
 
 	if mtd.IsClientStreaming() && mtd.IsServerStreaming() {
-		return invokeBidi(ctx, cancel, stub, mtd, handler, requestData, req)
+		return invokeBidi(ctx, stub, mtd, handler, requestData, req)
 	} else if mtd.IsClientStreaming() {
 		return invokeClientStream(ctx, stub, mtd, handler, requestData, req)
 	} else if mtd.IsServerStreaming() {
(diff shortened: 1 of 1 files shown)


grpcurl.go
```diff
@@ -554,13 +554,12 @@ func invokeServerStream(ctx context.Context, stub grpcdynamic.Stub, md *desc.Met
...
 }
 
-func invokeBidi(ctx context.Context, cancel context.CancelFunc, stub grpcdynamic.Stub, md *desc.MethodDescriptor, handler InvocationEventHandler,
+func invokeBidi(ctx context.Context, stub grpcdynamic.Stub, md *desc.MethodDescriptor, handler InvocationEventHandler,
 	requestData RequestMessageSupplier, req proto.Message) error {
 
 	// invoke the RPC!
 	str, err := stub.InvokeRpcBidiStream(ctx, md)
 
-	// mutex protects access to handler and sendErr since we'll have two goroutines sharing them
```
jhump: Also removed this errant comment, which does not apply (must have been describing earlier iteration of the code).

PR #38, opened 2018-07-06
Point to server reflection docs

jadekler:
Service reflection has very poor SEO and unfortunately not great discoverability. Since this package is an entry-point for service reflection, this helps make the process of finding this information easier.

Sidenote: I'm excited for the talk on grpcurl at Gophercon, and I think this will help users who investigate the package for the first time after seeing the talk. :)

Sidenote 2: I'm not absolutely sold on plopping in a huge code block right there at the top of the readme. If you prefer this in a different place, let me know,

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ In fact, the way the tool transforms JSON request data into a binary encoded pro
 is using that very same schema. So, if the server you interact with does not support
 reflection, you will need to build "protoset" files that `grpcurl` can use.
 
-[Examples for how to set up server reflection can be found here.](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md#known-implementations.)
+[Examples for how to set up server reflection can be found here.](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md#known-implementations)
 
 This repo also provides a library package, `github.com/fullstorydev/grpcurl`, that has
 functions for simplifying the construction of other command-line tools that dynamically
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ In fact, the way the tool transforms JSON request data into a binary encoded pro
 is using that very same schema. So, if the server you interact with does not support
 reflection, you will need to build "protoset" files that `grpcurl` can use.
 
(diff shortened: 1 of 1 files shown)


jhump:
@jadekler, thanks for the patch, but I think this info really should be presented in some other way.

Ideally, we could link to the main documentation for reflection (such as [this](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md)). Unfortunately, that doc is not particularly helpful since it is (1) inaccurate/out-of-date and (2) doesn't describe *how* to set it up for various languages. (There is [another doc](https://github.com/grpc/grpc/blob/master/doc/server_reflection_tutorial.md) which shows how to use reflection with C++, but that is woefully inadequate.)

While this repo, `grpcurl`, is written in Go, I think any mention of server-side reflection support should not be limited to Go servers, since this tool is meant for interacting with *any* gRPC server. Perhaps links to a few READMEs for multiple server languages would suffice, instead of this block of Go code?

Also, I think any new info along these lines belongs below, like in a new section titled "Service Reflection", adjacent to existing sections titled "Proto Source Files" and "Protoset Files".

jadekler:
I agree with all the above. I'll talk to Kailash - who runs grpc.io - on Monday and see whether he has a suggestion for where to put definitive service reflection docs. If we don't find a great place, I'll amend this PR as you suggested to link to various language's documentation under a new section.

jadekler:
Work progresses on this: https://github.com/grpc/grpc/pull/16018. I will update this PR once that one is finished.

@jhump separately, myself and a few other grpc members would love to have a meeting with you about grpcurl. Would you be interested in chatting with us? If so, could you shoot me an email at deklerk at google dot com?

jadekler:
@jhump PTAL. We have:

- Written server reflection tutorials for each of the languages server reflection is written in.
- Created a meta area in the server reflection doc that points to all the tutorials.
- I've changed this PR to a link to the meta area, which should be a good hopping point to implementing server reflection in the user's respective languages.

jadekler:
Done!

PR #40, opened 2018-07-06
allow period at the start of a symbol

### LINKED ISSUE #39: Describing service inputs/outputs
Hi @jhump. Thanks for this great project. I noticed that there is documentation for reflection of services - which I'm able to use without any problems - but there's no documentation / samples around reflection of service inputs and outputs. Is that at all possible? If so, does one go about it? I naively tried `grpcurl describe <the value of InputType that the service description gave>` but this did not work.

Thanks!

jadekler:
When describing a service, InputType and OutputType symbols are presented
as .Foo.Bar. However, when describing a type, .Foo.Bar is not accepted;
it must be changed to Foo.Bar. This is both confusing and breaks copy
paste ability. This commit addresses the problem by optionally allowing a
period at the beginning of a symbol.

Fixes #39.

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -369,6 +369,10 @@ func main() {
 			symbols = svcs
 		}
 		for _, s := range symbols {
+			if s[0] == '.' {
+				s = s[1:]
+			}
+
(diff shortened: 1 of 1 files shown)


jhump:
Will merge after green build from Travis.

PR #48, opened 2018-09-13
support go modules

jhump:
- use Go 1.11 (with and without module support) in CI
- add go.mod and go.sum files

### DIFF
--- a/grpcurl_test.go
+++ b/grpcurl_test.go
@@ -50,7 +50,7 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic(err)
 	}
-	sourceProtoFiles, err = DescriptorSourceFromProtoFiles(nil, "../../../google.golang.org/grpc/interop/grpc_testing/test.proto")
+	sourceProtoFiles, err = DescriptorSourceFromProtoFiles(nil, "testing/test.proto")
 	if err != nil {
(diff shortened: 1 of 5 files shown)


grpcurl_test.go
```diff
@@ -50,7 +50,7 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic(err)
 	}
-	sourceProtoFiles, err = DescriptorSourceFromProtoFiles(nil, "../../../google.golang.org/grpc/interop/grpc_testing/test.proto")
+	sourceProtoFiles, err = DescriptorSourceFromProtoFiles(nil, "testing/test.proto")
```
jhump: When building with Go 1.11 with `GO111MODULE=on` environment set, we can no longer rely on the file layout of dependencies all under a single `GOPATH` directory, so this relative reference to the gRPC project doesn't work. To work around (since it's just for a test anyway), I've copied/forked this `test.proto` file into this repo.

jhump:
@ianrose14, @dragonsinth, my intent is to create a "v1.0.0" release for this repo after landing this. WDYT?

ianrose14:
> my intent is to create a "v1.0.0" release for this repo after landing this. WDYT?

SGTM!

PR #49, opened 2018-09-15
add goreleaser

### LINKED ISSUE #47: Release binaries
Hi! I love using grpcurl but installing it on new machines on first use can be pretty slow to download the dependencies and compile. Would it be possible to start releasing binaries?

Suggested approach I've used on my projects is to use `goreleaser` to build binaries for a few operating systems and architectures, then it will upload them to the releases page of this project. 

Is this something that could be added?

Thanks!

leighmcculloch:
Add a goreleaser configuration file that builds binaries for Linux,
MacOS and Windows, for 32bit (x86/386) and 64bit (x64/amd64). The
binaries will be archived into a tar.gz/zip file along with the LICENSE
file.

The `dist/` directory will be written to by goreleaser with the binaries
during the build process, so it's also been added to .gitignore.

Goreleaser can be installed with:
`go get github.com/goreleaser/goreleaser`

Or other methods for install can be found at:
https://goreleaser.com

To build all the binaries and release them to GitHub:
1. Tag the release. e.g. `git tag -a v1.0.0 -m 'First release'`
2. Generate a GitHub Personal Access Token. See https://github.com/settings/tokens.
3. Push the release to GitHub. e.g. `git push origin v1.0.0`
4. Run goreleaser. e.g.  `GITHUB_TOKEN=xxxxxxx... goreleaser`

Goreleaser will build the binaries, and upload the binaries to the
GitHub project's release page for the release that was tagged.

Goreleaser will also set the `main.version` variable with the version
being released, so I added a `-version` flag that will print:
`grpcurl v<version>`

Closes #47


cmd/grpcurl/.goreleaser.yml
```diff
@@ -0,0 +1,22 @@
...
+  format: tar.gz
+  format_overrides:
+    - goos: windows
+      format: zip
+  replacements:
+    amd64: x64
+    386: x86
+    darwin: macos
+  files:
+    - LICENSE
```
jhump: When I tried to run this locally, this failed because the `LICENSE` file is not in this directory. However, when I tried editing this to point to `../../LICENSE`, it generated a gnarly entry in the resulting TGZ that had `../LICENSE` as the path (e.g. it created a directory entry in the archive for `..`, which seems bad).

I have not yet tried to read through `goreleaser` docs to see how to fix this situation, but I think we need this configuration to run and produce a good archive before we merge this PR.

I also noticed that the TGZ has a directory inside named `grpcurl_<version>_<platform>_<arch>`, which seems quite verbose. Many binary distributions I've seen (IIRC) just have a folder named `grpcurl` (without the version, platform, and architecture). Is there a "best practice" that I don't know about for putting all of this information into the resulting directory after unzipping a TGZ?

cmd/grpcurl/.goreleaser.yml
```diff
@@ -0,0 +1,22 @@
...
+      - 386
+
+archive:
+  wrap_in_directory: true
+  format: tar.gz
+  format_overrides:
+    - goos: windows
+      format: zip
+  replacements:
+    amd64: x64
```
jhump: I was just looking at the `protoc` distribution, and they use `x86_64` as the architecture here and `x86_32` for `386`. Maybe do the same, for consistency with a closely related binary?

cmd/grpcurl/.goreleaser.yml
```diff
@@ -0,0 +1,22 @@
...
+    386: x86
+    darwin: macos
```
jhump: The `protoc` distribution uses `osx`. So same remark as above: maybe change for consistency with that project?

cmd/grpcurl/grpcurl.go
```diff
@@ -30,13 +30,17 @@ import (
 	"github.com/fullstorydev/grpcurl"
 )
 
+var version = "<not set>"
```
jhump: Maybe something a little more informative, like `dev build <no version set>`? And it would also be nice to update the `Makefile` so it uses `-ldflags` to set `main.version` to something like `dev build <git sha>` (though that's fine to omit from this PR and maybe just add a TODO; if you want to tackle in another branch, great, and if not I can address it).

cmd/grpcurl/.goreleaser.yml
leighmcculloch: Ah yes. I originally had the `.goreleaser.yml` in the root, then moved it to the subdirectory. I can change the config so there's no nested directory. Some binaries I've seen don't both with the LICENSE, and they just have the binary without the directory. The easiest way to install these then is:

```
curl https://... | tar xz -C /usr/local/bin
```

cmd/grpcurl/.goreleaser.yml
leighmcculloch: 💯 👍

cmd/grpcurl/grpcurl.go
leighmcculloch: 👍 Lets tackle it in this branch. I'll add it.

cmd/grpcurl/.goreleaser.yml
leighmcculloch: Adding the license can be nice if someone redistributes the binary, but practically for all of the users just downloading it to install immediately, it's an annoyance and it makes the install instructions slightly longer to extract only the binary from the tar.

leighmcculloch:
I've pushed some changes to this branch, but I want to add the `dev build <git sha>` and will circle back to do that. Still would appreciate thoughts on https://github.com/fullstorydev/grpcurl/pull/49#discussion_r218095904. Thanks for already reviewing it so quickly!

cmd/grpcurl/.goreleaser.yml
jhump: For now, I think we should keep the license in the distribution. I think it's simple enough to extract the binary into `/usr/local/bin`, for people that want to install it there.

In the future, I may look into adding a homebrew recipe, which will simplify it further for OS X users, and make it even easier to auto-install into that location.

leighmcculloch:
@jhump Ready for another 👀 from you 😄 .

jhump:
@leighmcculloch, promise I haven't forgotten. I will try to make time for further review later today.

Makefile
```diff
@@ -1,3 +1,5 @@
+dev_build_version=$(shell git describe --always --dirty)
```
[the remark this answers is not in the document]
leighmcculloch: I can add `--tags`, and on a tag it will show the tag, and then on a non-tag commit it will display `dev build v1.0.0-12-gc162649-dirty`. It will list the previous tag, number of commits away from it, and the commit hash. It will also add on `dirty` if the working directory is changed.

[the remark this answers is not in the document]
leighmcculloch: I added this in 2a6a598.

.goreleaser.yml
```diff
@@ -0,0 +1,22 @@
+builds:
+  - binary: grpcurl
+    main: ./cmd/grpcurl/grpcurl.go
```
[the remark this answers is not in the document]
leighmcculloch: Fixed in bfbcb03.

leighmcculloch:
@jhump Addressed your last round of feedback. I also changed the format of the version text that's injected so it has a `v` on the beginning of it for the goreleaser builds. That's in 0620c43.

.goreleaser.yml
```diff
@@ -0,0 +1,24 @@
...
+    main: ./cmd/grpcurl
+    goos:
+      - linux
+      - darwin
+      - windows
+    goarch:
+      - amd64
+      - 386
+    ldflags:
+      - -s -w -X main.version=v{{.Version}}
```
leighmcculloch: I copied this line from the documentation, and the `-s -w` are part of the default (see below). The main difference I made was to drop the fields we don't use and to add a `v` at the beginning of the version. I did the prefix here and not in the code because the code would print out `v` in front of the `dev build...` string which wouldn't look great.

The default value is:
```
-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}
```
https://goreleaser.com/customization/#Name%20Templates

Makefile
```diff
@@ -16,7 +18,12 @@ updatedeps:
 
 .PHONY: install
 install:
-	go install ./...
+	go install -ldflags '-X "main.version=dev build $(dev_build_version)"' ./...
+
+.PHONY: release
+release:
+	@GO111MODULE=off go get github.com/goreleaser/goreleaser
```
leighmcculloch: I also added `GO111MODULE=off` here to make sure the `go get` command doesn't mess with the `go.mod` file, otherwise `go get` will add goreleaser and it's dependencies to `go.mod`. There's a chance the behavior of `go get` might change in go1.12 so that we don't have to turn it off because of very active conversation on golang/go#27643.

PR #56, opened 2018-10-17
improve flag doc with go 1.10+

jhump:
In Go 1.10, flag doc strings get auto-indented if they consist of multiple lines. This is not the case for older versions of Go. So this change makes the `grpcurl` usage output prettier in Go 1.10 (and consistent with older versions).

It also adds a test to verify that flag usage output is right. Finally, it improves the indentation, to make the doc strings easier to read and scan with a glance.

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -1,4 +1,4 @@
-// Command grpcurl makes GRPC requests (a la cURL, but HTTP/2). It can use a supplied descriptor
+// Command grpcurl makes gRPC requests (a la cURL, but HTTP/2). It can use a supplied descriptor
 // file, protobuf sources, or service reflection to translate JSON or text request data into the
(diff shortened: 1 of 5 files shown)


cmd/grpcurl/grpcurl.go
```diff
@@ -509,17 +513,45 @@ is given then the descriptors for all exposed or known services are shown.
...
+				buf.WriteString(indent())
+			}
+			buf.WriteString(line)
+		}
+		if pos >= len(docString) {
+			break
+		}
+		docString = docString[pos+1:]
+	}
+	return buf.String()
```
dragonsinth: would strings.Split not make this much more concise?

jhump: Yes it would. Not sure why I formulated it this way.

jhump: Fixed

PR #58, opened 2018-10-19
some pre-factoring and small fixes

jhump:
* Organize into multiple files
* Make listing methods show fully-qualified names
* Address small feedback from #56 (trim then check if empty)

This change is really just moving stuff into new files, to break up the very large `grpcurl.go` file. There are two other small fixes in here, which I'll point out with comments.


cmd/grpcurl/grpcurl.go
```diff
@@ -566,10 +566,11 @@ func prettify(docString string) string {
 	// from each line in the doc string
 	j := 0
 	for _, part := range parts {
+		part = strings.TrimSpace(part)
```
jhump: This is the first actual change. This is in reply to a comment you made on a commit from #56.

grpcurl.go
```diff
@@ -361,365 +121,13 @@ func ListMethods(source DescriptorSource, serviceName string) ([]string, error)
 	} else {
 		methods := make([]string, 0, len(sd.GetMethods()))
 		for _, method := range sd.GetMethods() {
-			methods = append(methods, method.GetName())
+			methods = append(methods, method.GetFullyQualifiedName())
```
jhump: This is the second change. It makes `grpcurl list <service name>` print out fully-qualified method names. I feel like almost every time I use this feature of the tool, I copy+paste the method name to do a `grpcurl describe <method>` and then get an error because it needs to be the full name.

PR #98, opened 2019-04-05
Adds support for showing error details

### LINKED ISSUE #96: Proposal: Add details in error output
Hello there!
I would like to add the `grpc status details` in the error output of the CLI, I simply want to display the response as google designed it, see [here](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto) 

Currently the CLI print the error like this: (`cmd/grpcurl/grpcurl.go`)
```go
if h.Status.Code() != codes.OK {
	fmt.Fprintf(os.Stderr, "ERROR:\n  Code: %s\n  Message: %s\n", 
		h.Status.Code().String(), h.Status.Message())
	exit(1)
}
```
I propose to add the details:
```go
if h.Status.Code() != codes.OK {
	fmt.Fprintf(os.Stderr, "ERROR:\n  Code: %s\n  Message: %s\n Details: %+v\n", 
		h.Status.Code().String(), h.Status.Message(), h.Status.Details())
	exit(1)
}
```

I will be happy to do the PR, so what do you think?

P.S: I was wondering if we could add an option to output this in JSON too, since the CLI already outputs in JSON for OK responses.

jhump:
This adds support for showing error details, if the server contained any.

To better support printing of `google.protobuf.Any` messages (which are what the error details are), this also makes a few other changes:
1. Allows printing of unresolvable `Any` messages using an `"@value"` field in JSON output that has the base64-encoded embedded message data. This will apply to both the error details, but also to any `Any` message that is embedded in a response message.
2. Improves support for `-format text` to show expanded `Any` messages if possible. (Due to limitations in underlying `proto` package, this will usually *not* be possible. But this should greatly improve with v2 of the go protobuf API.)
3. Addresses a TODO in existing `AnyResolver` code to lazily fetch descriptors as needed instead of having to download all files eagerly.

Fixes #96.


jhump:
@tsnyder701, I know you probably don't have much context. I'm happy to chat about it. I did this late last week in response to a bug filed (#96).

It's not that easy to test locally: you'd have to spin up a server that sends back details in its error messages. (I can share a gist of what I ran if it's helpful.) While testing, I noticed that the output is kind of ugly in default JSON format due to a bug in the upstream protobuf stuff. So I also have another PR out to remedy that (https://github.com/golang/protobuf/pull/834).

tsnyder701:
I've started looking at this. Can you paste an example of the output from before and after?

jhump:
@tsnyder701, sample output below. Here's the command I'm running (against a test server that is running [this service](https://github.com/fullstorydev/grpchan/blob/master/grpchantesting/test_service.go#L18)):

```
grpcurl -plaintext -d @ \
    127.0.0.1:58552 grpchantesting.TestService/Unary <<EOM
{
  "code":11,
  "error_details":[
    {"@type":"foo.com/grpchantesting.Message","code":13},
    {"@type":"foo.com/google.protobuf.Timestamp","value":"2019-04-08T13:24:35Z"}
  ]
}
EOM
```

*Before*

```
ERROR:
  Code: OutOfRange
  Message: error
```

*After*

```
ERROR:
  Code: OutOfRange
  Message: error
  Details:
  1)	{
    	  "@type": "foo.com/grpchantesting.Message",
    	  "code": 13
    	}
  2)	{
    	  "@type": "foo.com/google.protobuf.Timestamp",
    	  "value": "2019-04-08T13:24:35Z"
    	}
```

I also checked that it works with the text format. Here's a sample command for that:

```
grpcurl -plaintext -format text -d @ \
    127.0.0.1:58552 grpchantesting.TestService/Unary <<EOM
code: 11
error_details: <
  [foo.com/google.protobuf.Timestamp]: <seconds: 1551401446>
>
error_details: <
  [foo.com/google.protobuf.StringValue]: <value: "foobar">
>
EOM
```

The "before" output is the same as the example above -- just the code and message. But the "after" output looks like so:

```
ERROR:
  Code: OutOfRange
  Message: error
  Details:
  1)	[foo.com/google.protobuf.Timestamp]: <
    	  seconds: 1551401446
    	>
  2)	[foo.com/google.protobuf.StringValue]: <
    	  value: "foobar"
    	>
```

Note that if you were to compile it yourself and run it, `grpcurl` may not actually accept that input for the text format. I'm running a binary that is also linked with a [protoreflect patch](https://github.com/jhump/protoreflect/pull/185), to add some support for parsing the "expanded `Any` syntax" in the text format. (Note, I could not find this syntax documented anywhere, just like the rest of the text format. I just looked at the source in the [golang/protobuf](https://github.com/golang/protobuf/blob/master/proto/text.go#L207) repo.)

PR #100, opened 2019-05-15
Update README.md

jewzaam:
Use -u in go get to update dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ brew install grpcurl
 ### From Source
 You can use the `go` tool to install `grpcurl`:
 ```shell
-go get github.com/fullstorydev/grpcurl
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,7 +57,7 @@ brew install grpcurl
 ### From Source
 You can use the `go` tool to install `grpcurl`:
 ```shell
-go get github.com/fullstorydev/grpcurl
+go get -u github.com/fullstorydev/grpcurl
```
jhump: Perhaps this should instead be a comment, suggesting people to use the `-u` flag for `go get` if they are not yet using Go modules?

jhump:
This was somewhat relevant with older releases of Go. However this is definitely not the way to build things with Go 1.13 and up, due to Go modules.

PR #102, opened 2019-05-20
add API to expose AnyResolver implementations backed by a DescriptorSource

jhump:
The `grpcui` repo links in this package, since it also does dynamic RPC. For proper handling of Any messages (including error details), it really wants to use the same `AnyResolver` implementations that this package uses. So this API would make that possible (w/out copy+pasta of the code).

### DIFF
--- a/format.go
+++ b/format.go
@@ -195,6 +195,22 @@ const (
 	FormatText = Format("text")
 )
 
+// AnyResolverFromDescriptorSource returns an AnyResolver that will search for
+// types using the given descriptor source.
+func AnyResolverFromDescriptorSource(source DescriptorSource) jsonpb.AnyResolver {
+	return &anyResolver{source: source}
+}
+
+// AnyResolverWithFallbackFromDescriptorSource returns an AnyResolver that will
+// search for types using the given descriptor source and then fallback to a
+// special message if the type is not found. The fallback type will render to
+// JSON with a "@type" property, just like an Any message, but also with a
+// custom "@value" property that includes the binary encoded payload.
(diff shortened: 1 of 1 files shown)


format.go
```diff
@@ -195,6 +195,22 @@ const (
...
+func AnyResolverFromDescriptorSource(source DescriptorSource) jsonpb.AnyResolver {
+	return &anyResolver{source: source}
+}
+
+// AnyResolverWithFallbackFromDescriptorSource returns an AnyResolver that will
+// search for types using the given descriptor source and then fallback to a
+// special message if the type is not found. The fallback type will render to
+// JSON with a "@type" property, just like an Any message, but also with a
+// custom "@value" property that includes the binary encoded payload.
+func AnyResolverWithFallbackFromDescriptorSource(source DescriptorSource) jsonpb.AnyResolver {
```
tsnyder701: How about `AnyResolverFromDescriptorSourceWithFallback`? Despite it's slightly more awkward construction, I think it might convey the meaning a little better. When I first read the name here I was thinking it was an `AnyResolver` with a fallback from the descriptor source rather than an `AnyResolverFromDescriptorSource` with a fallback.

tsnyder701: An alternative could be to have AnyResolverFromDescriptorSource take a `...Options` that could indicate `WithFallback`.

jhump: I'm hesitant to add the options pattern unless the case clearly suggests that future options may be added. Seems awkward to add that pattern to support exactly one option.

I'll rename as suggested to `AnyResolverFromDescriptorSourceWithFallback`.

PR #104, opened 2019-05-30
add official Dockerfile

jhump:
I tried to craft this so the resulting image is as small as possible. At the same time, trying to adhere to best practices and also make the thing as useful as possible, it does not run as a privileged user in the container and also uses the latest root certs from alpine (so that TLS connections work as long as the site's cert is a typical issuer).

I tested using these commands:
```bash
make docker
docker run fullstorydev/grpcurl:v1.3.0-1-g80b118f --help
docker run fullstorydev/grpcurl:v1.3.0-1-g80b118f api.grpc.me:443 list
```

Assuming this looks good, my plan was to use it to push a v1.3.0 image to Docker Hub and to also add a `docker push` command to the `make release` target at some point.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,26 @@
 FROM golang:1.11.10-alpine as builder
 MAINTAINER FullStory Engineering
 
+# currently, a module build requires gcc (so Go tool can build
+# module-aware versions of std library; it ships only w/ the
+# non-module versions)
 RUN apk update && apk add --no-cache ca-certificates git gcc g++ libc-dev
 # create non-privileged group and user
 RUN addgroup -S grpcurl && adduser -S grpcurl -G grpcurl
 
-# build grpcurl
 WORKDIR /tmp/fullstorydev/grpcurl
 # copy just the files/sources we need to build grpcurl
 COPY VERSION *.go go.* /tmp/fullstorydev/grpcurl/
 COPY cmd /tmp/fullstorydev/grpcurl/cmd
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o /grpcurl -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" ./cmd/grpcurl
+# and build a completely static binary (so we can use
+# scratch as basis for the final image)
+ENV CGO_ENABLED=0
+ENV GOOS=linux
+ENV GOARCH=amd64
+ENV GO111MODULE=on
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN apk update && apk add --no-cache ca-certificates git gcc g++ libc-dev
+# create non-privileged group and user
+RUN addgroup -S grpcurl && adduser -S grpcurl -G grpcurl
+
+# build grpcurl
+WORKDIR /tmp/fullstorydev/grpcurl
+# copy just the files/sources we need to build grpcurl
+COPY VERSION *.go go.* /tmp/fullstorydev/grpcurl/
+COPY cmd /tmp/fullstorydev/grpcurl/cmd
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o /grpcurl -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" ./cmd/grpcurl
```
jhump: I got this from a website that provided tips for getting a binary that would run in a `scratch` container (which doesn't even have `libc`). The relevant tidbits are the linker flags and the `CGO_ENABLED=0`, so it doesn't try to use any other shared libraries for parts of the standard runtime.

jhump:
@sgmitchell, I did finally figure out the Docker issue I asked you about. I was missing the `CGO_ENABLE=0` part, to make sure the standard library wouldn't try to link against/load shared system libraries that are not part of scratch.

Dockerfile
sgmitchell: SGTM.

nit: readability could probably be improved by either adding those flag as env vars or using `\` to break up the lines
```
ENV CGO_ENABLED=0
ENV GOOS=linux
ENV GOARCH=amd64
ENV GO111MODULE=on 
RUN go build -o /grpcurl -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" ./cmd/grpcurl
```

```
RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on \
    go build -o /grpcurl -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" ./cmd/grpcurl
```

jhump:
I'm going to hold off merging this until I can get the fullstorydev org setup in Docker Hub, so I can go ahead and publish an image for v1.3.0 at the same time as merging the source for its Dockerfile.

PR #107, opened 2019-07-01
adding describe with json output

nadavwe:
hey @jhump,
I wanted to use grpcurl from another program and making the output "binary" way makes it much easier to consume.
I've added a flag to change the output of the describe command to be json instead of text.
the original behaviour stayed just as it was.
It's my first time writing in go, I hope I didn't make too much mess....
Thanks!

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


nadavwe:
hey @jhump,
basically we have an internal swiss-army-knife tool written in python and I wanted to have grpc capabilities in it. I thought of doing it on top of grpcurl - and for that I need some efficient way of passing data between the processes. 
The reason I found your tool is because there are not many tools to call arbitrary grpc services.
do you happen to know such python library? if there is then we can probably close this PR.

Thanks!

nadavwe:
ok, I delved more into the subject and I see I don't need the integration with grpcurl.
I'll just have it written in python using google.protobuf.
Thanks anyway, I was fun writing in go! :)

PR #111, opened 2019-08-09
 Switch from native flags to posix flags

andrewbenton:
This is beneficial because it makes grpcurl more like curl.  Many more
flags are now common between grpcurl and curl which will help users that
are new to the tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,10 +127,10 @@ source or protoset files.
 grpcurl localhost:8787 list
 
 # Using compiled protoset files
-grpcurl -protoset my-protos.bin list
+grpcurl --protoset my-protos.bin list
 
 # Using proto sources
-grpcurl -import-path ../protos -proto my-stuff.proto list
+grpcurl --import-path ../protos --proto my-stuff.proto list
 ```
 
 The "list" verb also lets you see all methods in a particular service:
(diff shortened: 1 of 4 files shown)


jhump:
@andrewbenton, does this _require_ long flag names to use two dashes and only allows short flag names with a single dash? If so, this is not backwards-compatible and will be a serious breaking change for anyone that has scripts that reference this tool.

I know the flag-handling is awkward due to the standard Go `flag` package, but I don't want to replace it with something that won't be backwards-compatible. My personal preference is just to swap for an implementation that allows flag aliases (e.g. multiple names for the same flag -- to support and short and long names) and that allows the arguments to be mixed in with flags, vs. having to come at the end. That would likely make it "close enough" to `curl`, but also preserve backwards-compatibility, especially since the existing flag support allows either one or two dashes for each flag. So users that were used to two dashes for long names and one dash for short names could still be accommodated.

andrewbenton:
There might be some other library that's able to accept the single dash as well as the double dashed form of the argument.  I'll have to see what I can find.  Under posix flags, something like `-proto` is actually treated as all of the single letter flags `p`, `r`, `o`, `t`, and `o` again.  Given that, we need a library that isn't posix flags if we want to keep that backward compatibility.

PR #115, opened 2019-09-13
it should use insecure flag instead of plaintext

xumc:
I use `-insecure` but it still give me a tls error message.

```
grpcurl -insecure localhost:8081 list
Failed to dial target host "localhost:8081": tls: first record does not look like a TLS handshake
```

After investigation, I found it seem we should use insecure flag instead of plaintext flag when we handle credentials.
After I fixed it, it works for me.

```
grpcurl -insecure localhost:8081 list
grpc.reflection.v1alpha.ServerReflection
proto.GrpcUpstreamService
```

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -290,7 +290,7 @@ func main() {
 			opts = append(opts, grpc.WithAuthority(*authority))
 		}
 		var creds credentials.TransportCredentials
-		if !*plaintext {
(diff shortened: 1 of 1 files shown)


jhump:
This is not right. The `-insecure` flag means you _are_ using TLS, but in an insecure way, ignoring the server's cert.

You need to actually use the `-plaintext` flag for this to work for your case.

xumc:
thanks

Issue #116, opened 2019-09-20
Headers can be read from stdin?


jammerful:
I have a use case where I'd the headers to be able to be read from stdin, in particular I'm passing credentials through the headers and I don't want the credentials to show up as command line parameters which exposes the credentials.

Would it be possible to allow the headers to be read from stdin like the request data? _I could put in a PR if this is simple, would you accept it?_

I was thinking of putting a switch like this for reading from stdin just like the request data: https://github.com/fullstorydev/grpcurl/blob/master/cmd/grpcurl/grpcurl.go#L489

jhump:
I think that's reasonable. But it could conflict with reading the request body from stdin. I think maybe the logic should look like so:

1. If only headers are read from stdin, then read stdin, line-by-line. Each line must be a header in `name: value` form. Blank lines are ignored as are lines that start with a `#` (I'm a sucker for allowing comments if/when it's easy to support!). It probably needs to trim, too -- though maybe only trim from the left (to avoid leading whitespace in the header name, but allow trailing whitespace in the value?).
2. If only the request is read from stdin, then the current behavior should be preserved.
3. If _both_ are to be read from stdin, it should have the headers first, one per line, and then a blank line, and then the request body (much like an HTTP 1.1 request might look).

What do you think? I doubt I'll have time to implement this soon, but I'll accept a patch if you're up for this.

jhump:
Actually, I just had a thought: what if instead we added a flag that allowed header values to be re-written via environment variable interpolation?

So instead of having to feed the headers in via stdin, you could do something like so:

```
grpcurl -expand-headers -H 'name: ${value}'
```

Note use of single quotes, so the shell won't expand `${value}`. Instead, grpcurl could do that, by search for `${...}` sequences and replacing them with the corresponding environment variable.

What do you think about that? I can actually see cases where both would be useful -- ability to interpolate env vars into the values and ability to pass via stdin... Anyhow, I think I'd take a patch that had either (or both!) of these approaches implemented.

jammerful:
Thanks for the second solution, I think that is much simpler and I'll put a PR in for that.

jammerful:
After putting in the PR, I would need a new release cut and build artifacts published. Is this possible?

jammerful:
@jhump Thanks for merging my PR, is there anyway you can cut a new release with accompanying binaries?

jhump:
@jammerful, we can leave this issue open as an enhancement request to also accept via stdin.

[...]

PR #117, opened 2019-09-24
Add Expand Headers Feature

jammerful:
See discussion of the feature here:
https://github.com/fullstorydev/grpcurl/issues/116

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -59,12 +59,13 @@ var (
 	rpcHeaders    multiString
 	reflHeaders   multiString
 	expandHeaders = flags.Bool("expand-headers", false, prettify(`
-		If set any environmental variables contained contained in any additional, rpc,
-		or reflection header string will be substituted for by the value of the environmental
-		variable. For instance, for the header 'key: ${VALUE}' where VALUE="foo"
-		will be expanded to 'key: foo'. Any undefined environmental variables
-		will throw an error. Note that verbose mode shows that the headers are
-		being expanded to. Escaping '${' is not supported.`))
+		If set, headers may use '${NAME}' syntax to reference environment variables.
+		These will be expanded to the actual environment variable value before
+		sending to the server. For example, if there is an environment variable
+		defined like FOO=bar, then a header of 'key: ${FOO}' would expand to 'key: bar'.
+		This applies to -H, -rpc-header, and -reflect-header options. No other
+		expansion/escaping is performed. This can be used to supply
+		credentials/secrets without having to put them in command-line arguments.`))
 	authority = flags.String("authority", "", prettify(`
 		Value of :authority pseudo-header to be use with underlying HTTP/2
 		requests. It defaults to the given address.`))
@@ -324,15 +325,15 @@ func main() {
 		var err error
 		addlHeaders, err = grpcurl.ExpandHeaders(addlHeaders)
 		if err != nil {
-			fail(err, "Failed to expand additional headers, missing environmental variable")
+			fail(err, "Failed to expand additional headers")
 		}
 		rpcHeaders, err = grpcurl.ExpandHeaders(rpcHeaders)
 		if err != nil {
-			fail(err, "Failed to expand rpc headers, missing environmental variable")
+			fail(err, "Failed to expand rpc headers")
 		}
 		reflHeaders, err = grpcurl.ExpandHeaders(reflHeaders)
 		if err != nil {
-			fail(err, "Failed to expand reflection headers, missing environmental variable")
+			fail(err, "Failed to expand reflection headers")
 		}
 	}
 
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -52,13 +52,19 @@ var (
 	key = flags.String("key", "", prettify(`
 		File containing client private key, to present to the server. Not valid
 		with -plaintext option. Must also provide -cert option.`))
-	protoset    multiString
-	protoFiles  multiString
-	importPaths multiString
-	addlHeaders multiString
-	rpcHeaders  multiString
-	reflHeaders multiString
-	authority   = flags.String("authority", "", prettify(`
+	protoset      multiString
+	protoFiles    multiString
+	importPaths   multiString
+	addlHeaders   multiString
+	rpcHeaders    multiString
+	reflHeaders   multiString
+	expandHeaders = flags.Bool("expand-headers", false, prettify(`
+		If set any environmental variables contained contained in the
+		header string will be substituted for by its corresponding environmental
+		variable. For instance, for the header 'key: ${VALUE}' where VALUE="foo"
+		will be evaluated to 'key: foo'. Note if no corresponding environmental
+		variable is found the header will be unchanged.`))
+	authority = flags.String("authority", "", prettify(`
 		Value of :authority pseudo-header to be use with underlying HTTP/2
 		requests. It defaults to the given address.`))
 	data = flags.String("d", "", prettify(`
@@ -313,6 +319,13 @@ func main() {
 		return cc
 	}
 
+	var headers []string
+	if *expandHeaders {
+		headers = grpcurl.ExpandHeaders(addlHeaders)
+	} else {
+		headers = addlHeaders
+	}
+
 	var cc *grpc.ClientConn
 	var descSource grpcurl.DescriptorSource
 	var refClient *grpcreflect.Client
@@ -329,7 +342,7 @@ func main() {
 			fail(err, "Failed to process proto source files.")
 		}
 	} else {
-		md := grpcurl.MetadataFromHeaders(append(addlHeaders, reflHeaders...))
+		md := grpcurl.MetadataFromHeaders(append(headers, reflHeaders...))
 		refCtx := metadata.NewOutgoingContext(ctx, md)
 		cc = dial()
 		refClient = grpcreflect.NewClient(refCtx, reflectpb.NewServerReflectionClient(cc))
@@ -502,7 +515,7 @@ func main() {
 		}
 		h := grpcurl.NewDefaultEventHandler(os.Stdout, descSource, formatter, *verbose)
 
-		err = grpcurl.InvokeRPC(ctx, descSource, cc, symbol, append(addlHeaders, rpcHeaders...), h, rf.Next)
+		err = grpcurl.InvokeRPC(ctx, descSource, cc, symbol, append(headers, rpcHeaders...), h, rf.Next)
 		if err != nil {
 			fail(err, "Error invoking method %q", symbol)
 		}
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -59,11 +59,12 @@ var (
 	rpcHeaders    multiString
 	reflHeaders   multiString
 	expandHeaders = flags.Bool("expand-headers", false, prettify(`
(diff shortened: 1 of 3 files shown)


jammerful:
@jhump As discussed here is the PR. Note I've made a few decisions regarding the functionality:

1. If no corresponding environmental variable is found, the header is sent as is. This is instead of an error being thrown causing grpcurl to terminate.
2. I've allowed header keys also to be set using environmental variables.

Open to changing the above.

grpcurl.go
```diff
@@ -161,6 +163,37 @@ func MetadataFromHeaders(headers []string) metadata.MD {
 	return md
 }
 
+/* Expands environmental variables contained in the header string
```
jammerful: Note I went with this comment format, as golint was complaining.

cmd/grpcurl/grpcurl.go
```diff
@@ -313,6 +319,13 @@ func main() {
 		return cc
 	}
 
+	var headers []string
+	if *expandHeaders {
+		headers = grpcurl.ExpandHeaders(addlHeaders)
```
jhump: Also need to expand `rpcHeaders` and `reflHeaders`.

grpcurl.go
jhump: Please stick with `//`, for consistency in this file.

To appease go lint, you need the comment to start with the named of the doc'ed thing, in this case "Expand Headers":

```
// ExpandHeaders expands environment variable references contained in the string.
// etc...
```

grpcurl.go
```diff
@@ -161,6 +163,37 @@ func MetadataFromHeaders(headers []string) metadata.MD {
...
+ * If no corresponding environmental variable is found, the header
+ * string is not changed. Hence, if the regex matches accidentally
+ * no changes are made.
+ */
+func ExpandHeaders(headers []string) []string {
+	expandedHeaders := make([]string, len(headers))
+	for idx, header := range headers {
+		if header != "" {
+			regex := regexp.MustCompile(`\${\w+}`)
```
jhump: Definitely don't compile regex's in the loop. I know it's a simple and fast tool, so performance at this step may not usually be a concern. But we can at least hoist this out and only compile it once (make it an unexported package `var`) in case anyone ends up using this API in an unexpected way in the future.

grpcurl.go
```diff
@@ -161,6 +163,37 @@ func MetadataFromHeaders(headers []string) metadata.MD {
...
+			results := regex.FindAllString(header, -1)
+			if results != nil {
+				expandedHeader := header
+				for _, result := range results {
+					envVarValue := os.Getenv(result[2 : len(result)-1])
+					replacementValue := envVarValue
+					// If no corresponding env var is found, leave the header as is.
+					if len(envVarValue) == 0 {
```
jhump: I think this will be confusing. Based on how other tools generally work with env vars, I would expect either an error that my input is bad or missing env vars to interpolate the empty string.

I'm a bit partial to the former since it is more likely catch bugs.

Although this does bring up an interesting point: if you want to use a literal `${` in the value, this scheme is too simple and would result in an error. I think we need a way to escape the `$`. For example, `$$` or `\$` meaning literal dollar sign, not the start of an env var reference.

jhump: Escaping can probably be left as a TODO since, by default, grpcurl does not try to do any variable expansion. So you can either do variable expansion or use literal dollar signs, which is likely sufficient for now.

jammerful:
Thanks for the great feedback, I will update shortly.

grpcurl.go
jammerful: Done.

cmd/grpcurl/grpcurl.go
jammerful: Done.

grpcurl.go
jammerful: Done. Sorry for oversight.

grpcurl.go
jammerful: Agreed, on the confusing behavior. Fixed.

cmd/grpcurl/grpcurl.go
```diff
@@ -313,6 +320,22 @@ func main() {
 		return cc
 	}
 
+	if *expandHeaders {
+		var err error
+		addlHeaders, err = grpcurl.ExpandHeaders(addlHeaders)
```
jammerful: Let me know if you wanted this DRY'd, was on the fence.

cmd/grpcurl/grpcurl.go
```diff
@@ -52,13 +52,20 @@ var (
...
-	reflHeaders multiString
-	authority   = flags.String("authority", "", prettify(`
+	protoset      multiString
+	protoFiles    multiString
+	importPaths   multiString
+	addlHeaders   multiString
+	rpcHeaders    multiString
+	reflHeaders   multiString
+	expandHeaders = flags.Bool("expand-headers", false, prettify(`
+		If set any environmental variables contained contained in any additional, rpc,
```
jhump: I think some of the wording in here is a little awkward. How about something like this instead:

```
If set, headers may use '${NAME}' syntax to reference environment variables. These
will be expanded to the actual environment variable value before sending to the server.
For example, if there is an environment variable defined like FOO=bar, then a header of
'key: ${FOO}' would expand to 'key: bar'. This applies to -H, -rpc-header, and
-reflect-header options. No other expansion/escaping is performed. This can be used
to supply credentials/secrets without having to put them in command-line arguments.
```

cmd/grpcurl/grpcurl.go
```diff
@@ -313,6 +320,22 @@ func main() {
 		return cc
 	}
 
+	if *expandHeaders {
+		var err error
+		addlHeaders, err = grpcurl.ExpandHeaders(addlHeaders)
+		if err != nil {
+			fail(err, "Failed to expand additional headers, missing environmental variable")
```
jhump: nit: "environment variable" instead of "environmental variable"

grpcurl.go
```diff
@@ -161,6 +163,36 @@ func MetadataFromHeaders(headers []string) metadata.MD {
 	return md
 }
 
+var envVarRegex = regexp.MustCompile(`\${\w+}`)
+
+// ExpandHeaders expands environmental variables contained in the header string.
+// If no corresponding environmental variable is found an error is thrown.
```
jhump: nit: "environment" instead of "environmental". Also "an error is returned" (not thrown).

grpcurl.go
```diff
@@ -161,6 +163,36 @@ func MetadataFromHeaders(headers []string) metadata.MD {
...
+	for idx, header := range headers {
+		if header != "" {
+			results := envVarRegex.FindAllString(header, -1)
+			if results != nil {
+				expandedHeader := header
+				for _, result := range results {
+					envVarName := result[2 : len(result)-1]
+					envVarValue := os.Getenv(envVarName)
+					replacementValue := envVarValue
+					if len(envVarValue) == 0 {
```
jhump: I think this is best tested using `os.LookupEnv` instead of `os.GetEnv`. That way, an explicitly blank environment variable is allowed -- but an _absent_ variable will result in an error:

```go
envVarValue, ok := os.LookupEnv(envVarName)
if !ok {
  return nil, fmt.Errorf("no variable %q is present in environment", envVarName)
}
```

grpcurl_test.go
```diff
@@ -300,6 +300,32 @@ func TestGetAllFiles(t *testing.T) {
...
+	expectedHeaders := map[string]bool{"": true, "key1: value": true, "key2: bar": true, "key3: ${woo": true, "key4: woo}": true,
+		"key5: value5": true, "key6: value6": true, "value5: value6": true}
+
+	outHeaders, err := ExpandHeaders(inHeaders)
+	if err != nil {
+		t.Errorf("The ExpandHeaders function generated an unexpected error %s", err)
+	}
+	for _, expandedHeader := range outHeaders {
+		if _, ok := expectedHeaders[expandedHeader]; !ok {
+			t.Errorf("The ExpandHeaders function has generated an unexpected header. Recieved unexpected header %s", expandedHeader)
```
jhump: nit: "Received" not "Recieved"

grpcurl_test.go
```diff
@@ -300,6 +300,32 @@ func TestGetAllFiles(t *testing.T) {
...
+		}
+	}
+
+	badHeaders := []string{"key: ${DNE}"}
+	_, err = ExpandHeaders(badHeaders)
+	if err == nil {
+		t.Errorf("The ExpandHeaders function should generate an error for missing environmental variables %q", badHeaders)
```
jhump: nit "environment" not "environmental"

grpcurl.go
```diff
@@ -161,6 +163,36 @@ func MetadataFromHeaders(headers []string) metadata.MD {
...
+
+// ExpandHeaders expands environmental variables contained in the header string.
+// If no corresponding environmental variable is found an error is thrown.
+// TODO: Add escaping for `${`
+func ExpandHeaders(headers []string) ([]string, error) {
+	expandedHeaders := make([]string, len(headers))
+	for idx, header := range headers {
+		if header != "" {
+			results := envVarRegex.FindAllString(header, -1)
+			if results != nil {
```
jhump: It is almost always best to check `len(results) > 0` instead of testing against nil (nil and empty slices should be treated the same).

grpcurl.go
```diff
@@ -161,6 +163,36 @@ func MetadataFromHeaders(headers []string) metadata.MD {
...
 
+var envVarRegex = regexp.MustCompile(`\${\w+}`)
+
+// ExpandHeaders expands environmental variables contained in the header string.
+// If no corresponding environmental variable is found an error is thrown.
+// TODO: Add escaping for `${`
+func ExpandHeaders(headers []string) ([]string, error) {
+	expandedHeaders := make([]string, len(headers))
+	for idx, header := range headers {
+		if header != "" {
```
jhump: Instead of deepening the logic with indentation under an `if` block, how about invert the condition:

```go
if header == "" {
    continue
}
```

Similar for below:

```go
if len(results) == 0 {
    expandedHeaders[idx] = headers[idx]
    continue
}

```

grpcurl.go
```diff
@@ -161,6 +163,36 @@ func MetadataFromHeaders(headers []string) metadata.MD {
...
+			results := envVarRegex.FindAllString(header, -1)
+			if results != nil {
+				expandedHeader := header
+				for _, result := range results {
+					envVarName := result[2 : len(result)-1]
```
jhump: Maybe a comment to clarify that these indices are to cut out the '${' in the beginning and the '}' at the end.

cmd/grpcurl/grpcurl.go
jhump: This seems fine.

cmd/grpcurl/grpcurl.go
jhump: Actually, the error returned from `grpcurl.ExpandHeaders` should suffice so that you don't need the "missing environment variable" bit -- it just makes the error that gets printed unnecessarily longer.

grpcurl.go
```diff
@@ -161,6 +163,36 @@ func MetadataFromHeaders(headers []string) metadata.MD {
...
+					envVarValue := os.Getenv(envVarName)
+					replacementValue := envVarValue
+					if len(envVarValue) == 0 {
+						return nil, fmt.Errorf("environmental variable '%s' is not set", envVarName)
```
jhump: It would be nice to reference the actual bad input. Maybe instead this:

```go
return nil, fmt.Errorf("header %q refers to missing environment variable %q", header, envVarName)
```

PR #120, opened 2019-09-30
add -protoset-out option

### LINKED ISSUE #118: Store protoset from reflection endpoint
It would be great if grpcurl would have a _switch_ that would allow to store protoset (and/or proto sources) to a file/directory. Maybe it could be a new _flag_ that would work with `describe` command, or a completely new command.

Our use-case: We want to automatically generate Envoy configuration with [gRPC-JSON transcoder](https://www.envoyproxy.io/docs/envoy/v1.9.0/configuration/http_filters/grpc_json_transcoder_filter) - it requires the protoset and list of services. grpcurl allows to list available services, just the protoset _export_ is missing.

jhump:
Resolves #118

This allows use of grpcurl to save protoset files of the descriptors that it resolves using service reflection. When doing `grpcurl <endpoint> list`, this will save a protoset file that has the definitions for _all_ of the services that the given endpoint exposes (assuming it supports server reflection).


cmd/grpcurl/grpcurl.go
```diff
@@ -101,6 +102,13 @@ var (
 		will accept. If not specified, defaults to 4,194,304 (4 megabytes).`))
 	emitDefaults = flags.Bool("emit-defaults", false, prettify(`
 		Emit default values for JSON-encoded responses.`))
+	protosetOut = flags.String("protoset-out", "", prettify(`
+		The name of a file to be written that will contain a FileDescriptorSet
+		proto. With the list and describe verbs, the listed or described
+		elements, and their transitive dependencies, will be written to the
```
dragonsinth: I think you can remove both these commas?

desc_source.go
```diff
@@ -251,3 +252,53 @@ func reflectionSupport(err error) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to serialize file descriptor set: %v", err)
+	}
+	if _, err := out.Write(b); err != nil {
+		return fmt.Errorf("failed to write file descriptor set: %v", err)
+	}
+	return nil
+}
+
+func addFilesToSet(seen map[string]struct{}, fds *[]*descpb.FileDescriptorProto, fd *desc.FileDescriptor) {
```
dragonsinth: could follow more normal append pattern, e.g. `allFilesSlice = addFilesToSet(allFilesSlice, expandedFiles, fds[filename])

cmd/grpcurl/grpcurl.go
jhump: done

desc_source.go
jhump: done

PR #122, opened 2019-10-14
Allow @ to be used for -protoset

bufdotbuild:
This allows i.e.:

```
buf image build -o - | grpcurl -protoset @ ...
```

This has been a long-standing pain point for me personally. The alternative is to make a temporary file for your FileDescriptorSet and then pass the path for this temporary file to grpcurl. In this PR, I make it so that `@` cannot be used for both `-data` and `-protoset`.


bufdotbuild:
Closing - for future reference, `grpcurl -protoset <(buf image build -o -) ...` does what I wanted.

PR #124, opened 2019-11-26
Register gzip decoder

shabbyrobe:
Hello, we have a server that sends gzip compressed responses for certain RPCs. We were getting the following error:

    ERROR:
      Code: Unimplemented
      Message: grpc: Decompressor is not installed for grpc-encoding "gzip"

This PR adds the import to register the gzip compressor. Our RPCs work as expected with this included.

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -23,6 +23,9 @@ import (
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 	reflectpb "google.golang.org/grpc/reflection/grpc_reflection_v1alpha"
+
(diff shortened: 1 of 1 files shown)


ejona86:
This doesn't sound right. The server is notified of the client-supported compression methods via `grpc-accept-encoding`. grpc-go appears to [only include grpc-accept-encoding when the request is using compression](https://github.com/grpc/grpc-go/blob/b3dcc681293a029bc998fc5259252b35bfef4817/internal/transport/http2_client.go#L437), which isn't ideal but also wouldn't cause this error. This sounds like a bug in the server. What server was being used?

It would be good to get an issue filed against the grpc server implementation for this. (Feel free to CC me on it)

(It's not bad to have compression capability, but if there's a bug here it should be fixed.)

PR #130, opened 2020-01-23
use wrapped TransportCredentials instead of handling handshake in Dialer

### LINKED ISSUE #129: TLS requests are made over http not https
I'm not sure if I understand this correctly, but making secure requests with certificate provided seems to make a request over http.

Call:
`grpcurl -cert certs\rootCA.pem -key certs\rootCA-key.pem -proto services\protos\MyService.proto -H "service: auth" -d "{\"Username\": \"test\", \"Password\": \"123\"}" grpc.local.dev:443 MyService.Signup`

Renders the following on the reverse proxy:
```
':method', 'POST'
':scheme', 'http'
':path', '/MyService/Signup'
':authority', 'grpc.local.dev:443'
'content-type', 'application/grpc' 
'user-agent', 'grpc-go/1.20.0-dev'
'te', 'trailers'
'service', 'auth'
```

This seems to be grpcurl problem as C# client sends request with the correct :scheme.

Please advice if I could overcome this somehow or if a fix is needed.

jhump:
This is so that the underlying gRPC library knows whether TLS is in use, so it will use the correct ":scheme" header.

Fixes #129.

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -71,8 +71,11 @@ var (
 		performed. This can be used to supply credentials/secrets without having
 		to put them in command-line arguments.`))
 	authority = flags.String("authority", "", prettify(`
-		Value of :authority pseudo-header to be use with underlying HTTP/2
-		requests. It defaults to the given address.`))
+		The authoritative name of the remote server. This value is passed as the
+		value of the ":authority" pseudo-header in the HTTP/2 protocol. When TLS
+		is used, this will also be used as the server name when verifying the
+		server's certificate. It defaults to the address that is provided in the
+		positional arguments.`))
 	data = flags.String("d", "", prettify(`
 		Data for request contents. If the value is '@' then the request contents
 		are read from stdin. For calls that accept a stream of requests, the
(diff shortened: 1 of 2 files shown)


jhump:
Welp, writing a test to actually inspect the ":scheme" header, to verify this fix, is not trivial. The existing TLS test fixtures would have to be refactored to support adding a proxy in the middle, to sniff the headers (because the Go gRPC libraries don't allow access to any of the reserved headers, such as ":scheme", ":method", etc).

So instead, I've manually verified (via breakpoints in the gRPC server library) that the patch works. I think that will suffice for now.

jhump:
By using `TransportCredentials`, this change inadvertently made `-authority` not work for TLS connections, because the dial option it was using was a plaintext-only option.

So I pushed another commit to restore it to work. However, this is not 100% backwards compatible: any usages that previously specified both `-servername` and `-authority` with _different_ values can no longer work. This is because the underlying Go gRPC libraries require these to be the same -- by making you configure them both with the same mechanism, which is the TLS server name.

So have to apply the `-authority` flag one way for plaintext and another way for TLS. And, for TLS, it means we're also specifying the TLS servername.

I update the help text for the flags to indicate this and also added some extra checks and warning messages to guide users in the right direction.

I don't _think_ this is a big enough change to warrant a v2 though. I think it's actually a security fix: after this patch, we won't trust a TLS server that can't provide a cert for the named authority. Any impacted users (hopefully zero, but almost certainly very few) would have to switch over to `-insecure` (or issue correct certs for their servers).

PR #136, opened 2020-03-12
Bump grpc to 1.28

gnossen:
@jhump  `grpcurl` is an amazing tool! Thank you for building and maintaining it!

gRPC just [released 1.28.0 with some pretty substantial changes](https://github.com/grpc/grpc-go/releases/tag/v1.28.0). This PR bumps the version of gRPC to 1.28.0.

Is there any chance we can get a new release of `grpcurl` with binaries?


gnossen:
[protobuf 1.3.3 depends on a Go version of at least 1.12](https://github.com/golang/protobuf/blob/d23c5127dc24889085f8ccea5c9d560a57a879d8/go.mod#L3). Per [the Go release policy](https://golang.org/doc/devel/release.html#policy) (two releases before the current release of 1.14), this version is unsupported anyway. My latest commit drops support for 1.11 and expands the test matrix to cover 1.13 and 1.14.

@jhump If you do not think this is the proper course of action, I'm happy to change course. Just let me know what you think is appropriate.

jhump:
> My latest commit drops support for 1.11 and expands the test matrix to cover 1.13 and 1.14.

Unfortunately, we can't take this change just yet. We (FullStory) are still migrating a bunch of code out of App Engine. That code is currently on the "go111" runtime, which is Go 1.11. So I don't want to accept any patches yet that would prevent us from staying on "latest" code in this repo, and we still have code that runs on Go 1.11. (I'm not sure when App Engine will drop support for 1.11, but they were supporting Go 1.9 well after the Go team officially stopped supporting it.)

I don't see any features in the latest gRPC that would actually provide any obvious improvement to the `grpcurl` command-line tool.

Also, since the build is green and you didn't have to actually touch any Go code, that means that you can of course use the `grpcurl` Go package in your own project, and use the very latest versions of protobuf, grpc, and Go.

So I don't think there's any urgency to upgrading. If you disagree, please elaborate.

go.mod
```diff
@@ -1,8 +1,10 @@
 module github.com/fullstorydev/grpcurl
 
+go 1.14
```
jhump: You need to remove this in order for it to work with Go v1.11.

gnossen: Sorry. Auto-added by `go mod tidy`. I suppose that's exactly how `protobuf` ended up with this erroneous dependency.

gnossen:
@jhump I talked with the go protobuf team. The dependency on `1.12` turned out to be an oversight. They've released a [`1.3.5`](https://github.com/golang/protobuf/releases/tag/v1.3.5) to address the issue, which I've pulled in instead. You're absolutely right that appengine should continue to be supported.

As for the importance of `1.28`, this release is the release of client-side support for the [`xDSv2` APIs](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api), which have been an area of major effort on the team for the past few months. We'd love to be able to take advantage of these in `grpcurl`.

go.sum
```diff
@@ -1,30 +1,62 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
jhump: There are a lot of changes here, and they are all additions -- no deletions. That is a little bit unsurprising.

Can you try locally deleting this file and then running `go test ./...`, to re-generate from scratch (and possibly prune some older no-longer-needed entries)?

gnossen: Nice catch. This caught some duplicate (unnecessary) dependencies in the file. Removed.

gnossen:
xDS usage does not require any code changes. It just requires an appropriate host string. You will specify, for example, `xds-experimental:///foohost:1234` to activate the xDS code paths.

jhump:
Do you not need an underscore import, to link in the package that actually registers the resolver for the `xds-experimental` scheme?

gnossen:
@jhump You're right about the underscore import. I'll follow on with one more PR to take care of that.

PR #137, opened 2020-03-12
Activate xDS support in grpcurl

gnossen:
Along with the version bump in https://github.com/fullstorydev/grpcurl/pull/136, this will activate xDSv2 support in [`grpcurl`](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api).

@jhump Sorry for the barrage of PRs. A release with binaries would be highly appreciated after this PR.

CC @dfawley

**Edit: This change has manually been verified against an xDS server.**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,17 +3,17 @@ sudo: false
 
 matrix:
   include:
-    - go: "1.9"
-    - go: "1.10"
-    - go: "1.11"
+    - go: 1.9.x
+    - go: 1.10.x
+    - go: 1.11.x
       env:
       - GO111MODULE=off
       - VET=1
-    - go: "1.11"
+    - go: 1.11.x
       env: GO111MODULE=on
-    - go: "1.12"
+    - go: 1.12.x
       env: GO111MODULE=off
-    - go: "1.12"
+    - go: 1.12.x
       env: GO111MODULE=on
     - go: tip
 
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -32,6 +32,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/metadata"
+	_ "google.golang.org/grpc/xds/experimental"
 )
 
 // ListServices uses the given descriptor source to return a sorted list of fully-qualified


gnossen:
I see the CI failure. Working on it.

gnossen:
@jhump This should be good to go now. I had to update the Travis configuration to use a newer patch to `1.11` than `1.11.1`, which comes pre-installed. Starting with `1.11.3`, the behavior with respect to the `go` statement in `go.mod` in libraries was changed to be treated as a recommendation. I went ahead and used the latest patch of each minor version test environment as well.

gnossen:
Thanks for the speedy review! 😀

jhump:
@gnossen, do you mind sending me a short blurb or some links on how to actually use the xDS stuff? The grpc-go release page only has a link to an Envoy API page. Clicking around Envoy docs, I also found [this](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol), but that is hella complicated and doesn't really tell me what I want to know.

Mainly, what does the actual address (after the `xds-experimental` scheme) look like? Is there HOWTO documentation that I could link to? (The Envoy page seems more like an xDS protocol primer.)

I basically need something to put in release notes in order create a new v1.5.0 release. And I haven't really been following xDS (other than knowing it's replacing the grpclb protocol, right?) and don't understand it well enough to describe it!

gnossen:
@jhump [This gRFC](https://github.com/grpc/proposal/pull/170) should tell you what you want to know about how gRPC integrates with xDS at a high level. At a code level, [here is an example](https://github.com/menghanl/grpc-go/commit/160c5ebc3ce9458c2d1690ec1c0541fc6e9bb34e) of what the scheme and bootstrap configuration file need to look like.

I think the following would be a decent write-up for the release notes:

> gRPC currently supports its own "grpclb" protocol for look-aside load-balancing. However, the popular Envoy proxy uses the xDS API for many types of configuration, including load balancing, and that API is evolving into a standard that will be used to configure a variety of data plane software. In order to converge with this industry trend, gRPC will be moving from its original grpclb protocol to the new xDS protocol.

> The xDS protocol allows configuring a variety of features, including many that gRPC does not currently support. Over time, the gRPC team will add support for more of these features. This release includes experimental support for client side xDSv2 APIs. Only the virtual host matching and cluster route action is supported. More features will be added in future.

> To use xDS, set up a bootstrap file as indicated in [the xDS gRFC](https://github.com/grpc/proposal/pull/170) and address a load-balanced name `example.com` with the string `xds-experimental///:example.com`. Note the triple slashes.

Feel free to edit as desired. Let me know if you need any more information.

Thank for all of your help!

gnossen:
@jhump Anything else you need on my end for a new release?

jhump:
@gnossen, I made a v1.5.0 release and also pushed latest image to docker hub. I hope it works how you want it. I was a bit brief in the release notes, so hopefully it makes sense: https://github.com/fullstorydev/grpcurl/releases/tag/v1.5.0

gnossen:
Perfect! Thanks for your help! `grpcurl` is an awesome tool. I've been trying to evangelize it wherever possible.

PR #144, opened 2020-04-01
drop 1.9 and 1.10; add 1.13 and 1.14

jhump:
1.9 and 1.10 are ancient and even App Engine no longer supports them.

1.13 and 1.14 are the new hotness.


jhump:
CI passed for both, but looks like the Travis webhook didn't make it to GitHub...

PR #147, opened 2020-04-06
cmd/grpcurl: add -allow-unknown-fields option

adambabik:
This PR adds a new option `-allow-unknown-fields` (its default value is `false` so the original behavior does not change) which, if present, will instruct JSON protobuf parser to ignore unknown fields present in the request.

My use case is that I have two services sharing a message. The message got extend with a new field but the services haven't been upgraded yet. The change to the message is backward compatible but I can't use `grpcurl` due to an error: `error getting request data: message type ABC has no known field named my_new_field`.

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -100,6 +100,10 @@ var (
 		ASCII character: 0x1E. The stream should not end in a record separator.
 		If it does, it will be interpreted as a final, blank message after the
 		separator.`))
+	allowUnknownFields = flags.Bool("allow-unknown-fields", false, prettify(`
+		When true, the request contents, if 'json' format is used, allows
+		unkown fields to be present. They will be ignored when parsing
+		the request.`))
 	connectTimeout = flags.Float64("connect-timeout", 0, prettify(`
 		The maximum time, in seconds, to wait for connection to be established.
 		Defaults to 10 seconds.`))
@@ -615,7 +619,8 @@ func main() {
 				// for messages, also show a template in JSON, to make it easier to
 				// create a request to invoke an RPC
 				tmpl := grpcurl.MakeTemplate(dsc)
-				_, formatter, err := grpcurl.RequestParserAndFormatterFor(grpcurl.Format(*format), descSource, true, false, nil)
+				options := grpcurl.FormatOptions{EmitJSONDefaultFields: true}
+				_, formatter, err := grpcurl.RequestParserAndFormatter(grpcurl.Format(*format), descSource, nil, options)
 				if err != nil {
 					fail(err, "Failed to construct formatter for %q", *format)
 				}
@@ -647,7 +652,12 @@ func main() {
 		// between each message, so output could potentially be piped
 		// to another grpcurl process
 		includeSeparators := !*verbose
-		rf, formatter, err := grpcurl.RequestParserAndFormatterFor(grpcurl.Format(*format), descSource, *emitDefaults, includeSeparators, in)
+		options := grpcurl.FormatOptions{
+			EmitJSONDefaultFields: *emitDefaults,
+			IncludeTextSeparator:  includeSeparators,
+			AllowUnknownFields:    *allowUnknownFields,
+		}
+		rf, formatter, err := grpcurl.RequestParserAndFormatter(grpcurl.Format(*format), descSource, in, options)
 		if err != nil {
 			fail(err, "Failed to construct request parser and formatter for %q", *format)
 		}
--- a/format.go
+++ b/format.go
@@ -55,6 +55,15 @@ func NewJSONRequestParser(in io.Reader, resolver jsonpb.AnyResolver) RequestPars
 	}
 }
 
+// NewJSONRequestParserWithUnmarshaler is like NewJSONRequestParser but
+// accepts a protobuf jsonpb.Unmarshaler instead of jsonpb.AnyResolver.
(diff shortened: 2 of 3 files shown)


adambabik:
@jhump thanks for checking. You're totally right, in this form the package itself requires a major version bump. I will reformulate this PR to avoid that.

adambabik:
@jhump sorry, it took me actually a while to find some spare time. I rewrote the code so that it does not alter the public interface.

format.go
```diff
@@ -342,23 +351,46 @@ func (a *unknownAny) ProtoMessage() {
 
 var _ proto.Message = (*unknownAny)(nil)
 
-// RequestParserAndFormatterFor returns a request parser and formatter for the
+// FormatterOptions is a set of flags that are passed to a JSON or text formatter.
+type FormatterOptions struct {
+	EmitJSONDefaultFields bool // FormatJSON only
+	IncludeTextSeparator  bool // FormatText only
+	AllowUnknownFields    bool // FormatJSON only
```
jhump, line 358: Let's group the JSON related fields together, instead of interleaving.

format.go
```diff
@@ -342,23 +351,46 @@ func (a *unknownAny) ProtoMessage() {
...
-// data (if needed by the format). The flags emitJSONDefaultFields and
-// includeTextSeparator are options for JSON and protobuf text formats,
-// respectively. Requests will be parsed from the given in.
-func RequestParserAndFormatterFor(format Format, descSource DescriptorSource, emitJSONDefaultFields, includeTextSeparator bool, in io.Reader) (RequestParser, Formatter, error) {
+// data (if needed by the format).
+// It accepts a set of options. The field EmitJSONDefaultFields and IncludeTextSeparator
+// are options for JSON and protobuf text formats, respectively. The AllowUnknownFields field
+// is a JSON-only format flag.
+// Requests will be parsed from the given in.
+func RequestParserAndFormatterWithOptions(format Format, descSource DescriptorSource, opts FormatterOptions, in io.Reader) (RequestParser, Formatter, error) {
```
jhump, line 368: Let's shorten this to `RequestParserAndFormatter` and then deprecate the other version.

Also, please move `opts` to be the last argument since that is more idiomatic for functions that use an options-struct pattern.

format.go
jhump: Technically `AllowUnknownFields` is an option for the parser, not the formatter.

How about rename this to `FormatOptions` -- since they can apply to parsing the format, not just outputting it.

format.go
```diff
@@ -342,23 +351,46 @@ func (a *unknownAny) ProtoMessage() {
...
 	}
 }
 
+// RequestParserAndFormatterFor returns a request parser and formatter for the
+// given format. The given descriptor source may be used for parsing message
+// data (if needed by the format). The flags emitJSONDefaultFields and
+// includeTextSeparator are options for JSON and protobuf text formats,
+// respectively. Requests will be parsed from the given in.
+func RequestParserAndFormatterFor(format Format, descSource DescriptorSource, emitJSONDefaultFields, includeTextSeparator bool, in io.Reader) (RequestParser, Formatter, error) {
+	opts := FormatterOptions{
```
jhump, line 387: After moving `opts` to end of arg list for subsequent call, this could be inlined.

cmd/grpcurl/grpcurl.go
```diff
@@ -615,7 +619,12 @@ func main() {
 				// for messages, also show a template in JSON, to make it easier to
 				// create a request to invoke an RPC
 				tmpl := grpcurl.MakeTemplate(dsc)
-				_, formatter, err := grpcurl.RequestParserAndFormatterFor(grpcurl.Format(*format), descSource, true, false, nil)
+				options := grpcurl.FormatterOptions{
+					EmitJSONDefaultFields: true,
+					IncludeTextSeparator:  false,
+					AllowUnknownFields:    *allowUnknownFields,
```
jhump, line 625: Let's just hardcode to `false`, since this code path won't be parsing anything.

Also, you can just omit the literal false values since they are default/zero vals:

```
options := grpcurl.FormatOptions{EmitJSONDefaultFields: true}
```

format.go
```diff
@@ -342,23 +351,60 @@ func (a *unknownAny) ProtoMessage() {
...
 	}
 }
 
+// RequestParserAndFormatterFor returns a request parser and formatter for the
+// given format. The given descriptor source may be used for parsing message
+// data (if needed by the format). The flags emitJSONDefaultFields and
+// includeTextSeparator are options for JSON and protobuf text formats,
+// respectively. Requests will be parsed from the given in.
+// This function is deprecated. Please use RequestParserAndFormatter instead.
+// DEPRECATED
```
adambabik, line 400: I am not sure what's project's convention to mark functions as deprecated. I explained it in the comment and added `// DEPRECATED` line. The latter works nicely with editors, for example GoLand, that inform when a deprecated function is used.

adambabik:
@jhump good stuff, thanks for a detailed review. I applied all suggestions, added a bit more code comments, and fixed tests to use `RequestParserAndFormatter` instead of now deprecated `RequestParserAndFormatterFor`.

format.go
jhump: https://github.com/golang/go/wiki/Deprecated

This is fine. I may tweak it in a later commit. More idiomatic for Go would be to take the last two lines here and change them to something like so:

```go
// Deprecated: Use RequestParserAndFormatter instead.
```

PR #153, opened 2020-04-23
format: fix crash when emitting empty messages in text format

q3k:
This fixes a `str[-1]` dereference if the marshaled text proto is an empty string.

There might be a better way to fix this - maybe `strings.TrimRight(str, "\n")`?


jhump:
Looks like the notification from Travis to GitHub failed. But the build succeeded and tests passed: https://travis-ci.org/github/fullstorydev/grpcurl/builds/678751210

PR #155, opened 2020-05-01
Add [-error-format (text|json)]  argument

mkatychev:
* This was added to help in testing grpcurl error output in its entirety. 
* Current error output doesn't follow any well known format making error deserialization/parsing a very ad-hoc process.

* added [`PrintFormattedStatus`](https://github.com/fullstorydev/grpcurl/blob/d2c9c318b71d3acd504fa5a47f576f453b595b39/format.go#L486-L494) to `format.go`  as an alternate status writer function:

```go
func PrintFormattedStatus(w io.Writer, stat *status.Status, formatter Formatter)
```

* if the  `-error-format json` flag is passed to grpcurl and response status code is non-zero, `Print` will output the protobuf `"google.golang.org/genproto/googleapis/rpc/status".Status` as JSON to stderr.


jhump:
I think the flag instead needs to be something about the error output being formatted. Like `-format-error`. When true, it skips the headings/labels and just prints the formatted status. The actual format, json vs. text, should still probably be driven via the existing format flag. So if JSON is being used for the response output, it is also used for the error output. (This is actually already the case -- the details are formatted in JSON vs text based on the selected format for the response. So I think we just want a flag to remove the ceremony intended for human readers.)

mkatychev:
* It felt like a bit of a code smell for me to use the formatter generated by functions such as `RequestParserAndFromatterFor` for use as the response formatter.

* [Getting the descriptor source  for an `*spb.Status` feels very cumbersome to get a particular any resolver](https://github.com/fullstorydev/grpcurl/blob/6a3c881d8ec6b5a4952f98f4e9835a297442f8c4/format.go#L353).
I ended up passing the basic JSON marshaller to `NewJSONFormatter`:
```go
formatter = NewJSONFormatter(
	emitJSONDefaultFields,
	anyResolverWithFallback{AnyResolver: jsonpb.Marshaler{Indent: "  "}.AnyResolver})
```
If there's a straightforward way to get the status resolver, I'd love to know.

jhump:
> It felt like a bit of a code smell for me to use the formatter generated by functions such as RequestParserAndFromatterFor for use as the response formatter.

I don't smell it :P
That seems to me like exactly what that formatter is for.

> If there's a straightforward way to get the status resolver, I'd love to know.

Just use the same `formatter` used to print response messages?

format.go
```diff
@@ -467,3 +467,13 @@ func PrintStatus(w io.Writer, stat *status.Status, formatter Formatter) {
 		}
 	}
 }
+
+// PrintFormattedStatus writes the gRPC status response in its entirety given the
+// provided Formatter
+func PrintFormattedStatus(w io.Writer, stat *status.Status, formatter Formatter) {
```
jhump, line 473: I don't think this function carries its weight: its not worth adding to the exported surface area of the `grpcurl` package. Let's just inline it into the `main` package. The added test can just be removed since it don't really do much beyond existing tests in `format_test.go` anyway.

mkatychev: Sounds good, I added right under `dial`.

cmd/grpcurl/grpcurl.go
```diff
@@ -652,7 +663,11 @@ func main() {
 			fmt.Printf("Sent %d request%s and received %d response%s\n", reqCount, reqSuffix, h.NumResponses, respSuffix)
 		}
 		if h.Status.Code() != codes.OK {
-			grpcurl.PrintStatus(os.Stderr, h.Status, formatter)
+			if *formatError {
+				printFormattedStatus(os.Stderr, h.Status, formatter)
+			} else {
+				grpcurl.PrintStatus(os.Stderr, h.Status, formatter)
+			}
 			exit(1)
```
mkatychev, line 671: Does it make sense for the `exit(1)` here to use the response status code instead: `exit(int(h.Status.Code()))`?

https://github.com/fullstorydev/grpcurl/blob/8243c87e864cbe2ad959d1e0ecdd26576258657b/cmd/grpcurl/grpcurl.go#L671

jhump: I like that idea!

But I think we should use `64+code` to avoid confusion as to whether between an exit code of 1 or 2 means `Canceled` or `Unknown` vs. some other error. This program already uses exit code 1 to mean unexpected error (like failures that prevented an RPC from even being sent, such as specifying bad filenames, being unable to dial server, etc). And we also use exit code 2 to mean incorrect usage (bad combination of command-line args).

So to avoid confusion with those scenarios, the range 64+ should be good to mean that RPC was actually performed but ultimately failed.

I suggest 64 as the constant (as opposed to say 4, 8, 16, or 32) since that is the range recommended by a resource I found on the web for user-defined exit codes: http://www.tldp.org/LDP/abs/html/exitcodes.html
Seems like reasonable advice...

jhump:
FYI, if you rebase against latest master, that build failure should go away.

PR #156, opened 2020-05-12
make sure we use pristine go.mod for test

jhump:
When we vet, it looks like one of the tools is pulling in protobuf v1.4 into `go.mod`, even though the checked-in version of `go.mod` clearly specifies 1.3.5 :(

Vetting with v1.4 fails because it marks everything imported as `github.com/golang/protobuf` as deprecated (which results in a vet error), but we aren't ready to move to the new `google.golang.org/protobuf` APIs yet.

This hack is an attempt to ensure that the test step uses the pristine `go.mod` file, instead of one that may have been modified by the Go tool when running the various vet tools.


jhump:
That's not gonna work...

PR #157, opened 2020-05-12
get travis CI green again

jhump:
* fix test to work w/ newer grpc repo
* use Go 1.13 in CI vet step, not Go 1.11
* run `make ci` to have it update `go.mod` and `go.sum`, so all tool dependencies are pinned

Green build: https://travis-ci.org/github/fullstorydev/grpcurl/builds/685941881


jhump:
Two things have gone wrong:
1) Seventeen days ago (!!) the descriptors linked into grpc had their canonical names changed. [First failed build](https://travis-ci.org/github/fullstorydev/grpcurl/builds/679399577)
2) Five days ago, something happened in Go 1.11 such that it's trying to pull in v1.4 of the protobuf runtime when resolving all dependencies of this module. But this module only depends on v1.3.5. The problem with v1.4 is that it deprecates a lot of functions used by grpcurl, which then makes staticcheck unhappy, causing a [red build](https://travis-ci.org/github/fullstorydev/grpcurl/jobs/684273489).

I could not repro that second point with a newer version of Go. So this PR moves the `VET` step to use Go 1.13 instead of Go 1.11. That fixes one of the red jobs. The other two are fixed by updating the test to allow the file names observed in the newer grpc version.

Frustratingly, I can't find any email from Travis from when this repo first failed, 17 days ago. (I usually try to fix red builds much more promptly than letting over two weeks elapse.) The last email I have from Travis about this repo is from 21 days ago (April 20th). I'll have to dig into that...

PR #159, opened 2020-05-14
fix broken build: use a rune literal, not an int literal

jhump:
Go tip does not like using `string(i)` where `i` is an untyped int
https://travis-ci.org/github/fullstorydev/grpcurl/jobs/686844524


jhump:
Green build: https://travis-ci.org/github/fullstorydev/grpcurl/builds/687125733

PR #160, opened 2020-05-15
use go 1.13 to build docker image

jhump:
With go 1.11, building with modules in a vanilla docker image required installing more junk.

With go1.13, mui simple.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,6 @@
 FROM golang:1.13.10-alpine3.11 as builder
 MAINTAINER FullStory Engineering
 
-# currently, a module build requires gcc (so Go tool can build
-# module-aware versions of std library; it ships only w/ the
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,6 @@
-FROM golang:1.11.10-alpine as builder
+FROM golang:1.13.10-alpine3.11 as builder
```
dragonsinth, line 1: hmm, i wouldn't have expected the trailing version number on `alpine`

jhump: It isn't strictly necessary. It's to pin the version. You can also specify just `golang:1.13-alpine`, to leave point release from Go and version entirely from alpine, and it will just pick up latest of each.

I thought including full versions would be good for repeatability -- always using the same version of Go and alpine to build the docker image.

PR #162, opened 2020-05-15
unpin Go point release and alpine version in Dockerfile

jhump:
After thinking about it more, maybe repeatable Docker builds aren't as important as not having to touch the Dockerfile as often to upgrade to newer Go point releases. I personally only ever run `docker build` when doing releases and pushing the images to Docker hub. I'm not sure how many users try to build their own docker file, but presumably, if they don't want the image in Docker hub, then maybe they're not trying to reproduce the exact bits in the release anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.13.10-alpine3.11 as builder
+FROM golang:1.13-alpine as builder
 MAINTAINER FullStory Engineering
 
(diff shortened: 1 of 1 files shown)


jhump:
> you can't build on current tip image I think I'd want CI to fail and let me know.

That would be a neat addition to CI. But currently, it doesn't try to build Docker images. (I only ever build images for releases.)

PR #170, opened 2020-07-14
move import of xds package to cmd/grpcurl; other go.mod and Makefile tweaks

jhump:
My main intent was to move the import of "google.golang.org/grpc/xds" out of the main API package and into the main command package (since it's the command that needs its `init` functions to have run, not the API).

But I ran into an issue testing: some deps kept auto-upgrading when running `make ci`. I realized it was because the make targets were relying on a pre-modules formulation -- using `go get`. But this causes the Go tool to retrieve the latest version, which would then change the version pinned in `go.mod`. So I changed all of these to `GO111MODULE=on go install`, so they just use the version already pinned.

I went ahead and left the deps upgraded though (for `staticcheck` and `predeclared` CI checkers).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ install:
 
 .PHONY: release
 release:
-	@go get github.com/goreleaser/goreleaser
+	@GO111MODULE=on go install github.com/goreleaser/goreleaser
 	goreleaser --rm-dist
 
 .PHONY: docker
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ install:
 
 .PHONY: release
 release:
-	@go get github.com/goreleaser/goreleaser
+	@GO111MODULE=on go install github.com/goreleaser/goreleaser
```
dragonsinth, line 25: maybe just turn this on for the entire makefile?

jhump: Well, we still run tests in travis w/out Go modules, to make sure that the project can be built either way. So turning it on in the Makefile would mean those CI jobs aren't doing what they're supposed to be doing.

At some point we can revisit dropping support for non-module builds. But that seemed more aggressive than the intended scope of this branch.

PR #181, opened 2020-08-01
Print size of the message in very verbose mode

vors:
Thank you for the great tool!

### Why

I found myself in a need of estimating serialized proto sizes few times.
Seems like a generally useful feature for `-v` mode.

### DIFF
--- a/format.go
+++ b/format.go
@@ -409,13 +409,13 @@ func RequestParserAndFormatterFor(format Format, descSource DescriptorSource, em
 // safe for use with InvokeRPC as long as NumResponses and Status are not read
 // until the call to InvokeRPC completes.
 type DefaultEventHandler struct {
-	out        io.Writer
-	descSource DescriptorSource
-	formatter  func(proto.Message) (string, error)
+	Out        io.Writer
+	DescSource DescriptorSource
+	Formatter  func(proto.Message) (string, error)
 	// 0 = default
 	// 1 = verbose
 	// 2 = very verbose
-	verbosityLevel int
+	VerbosityLevel int
 
 	// NumResponses is the number of responses that have been received.
 	NumResponses int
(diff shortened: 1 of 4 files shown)


format.go
```diff
@@ -459,6 +459,7 @@ func (h *DefaultEventHandler) OnReceiveHeaders(md metadata.MD) {
 func (h *DefaultEventHandler) OnReceiveResponse(resp proto.Message) {
 	h.NumResponses++
 	if h.verbose {
+		fmt.Fprintf(h.out, "\nResponse size (bytes): %d\n", proto.Size(resp))
```
jhump, line 462: This is not necessarily the serialized size of the message. This is the size of the _canonical_ serialized size. It is possible that the actual bytes received on the wire differ from this. In fact, the on the wire data could have even been compressed.

And, unfortunately, it's not as simple to capture that real size. It is easier to capture the decompressed size, prior to unmarshaling, via a `ForceCodec` call option. Although that would require more intrusive changes to this package. Getting at the actual on-the-wire size is harder due to how compressors are registered in the gRPC Go runtime. I think the `main` package would have to register a custom compressor that uses the same type/scheme, but then wraps the actual compressor implementation, just to intercept the actual on-the-wire byte slice, to measure its size.

vors:
I see, yeah I guess I wasn’t precise in my formulation. Canonical size is fine for me and still useful. The problem that without anything like that it’s hard to even estimate the size just by looking at the decoded text.

The context is that I wanted to estimate size for a caching layer.

However if you feel like embarking on making it more accurate that would be great.
What I can do is change the wording to “canonical size” and perhaps it’s still a meaningful change? Let me know if you’d like that.

jhump:
@vors, I had a thought that would make this "estimated size" more opt-in (so hopefully less likely to be misleading) and also address the red tests.

Let's add a `-vv` flag that means "very verbose". Under the hood, this can translate into a numeric verbosity setting: `-v` = 1, `-vv` = 2.

To make this change without having to add more factory function signatures (or adding an options pattern), for creating `DefaultEventHandler`, we can just make the `out`, `formatter`, and `verbosity` fields of that struct _exported_. (We can actually _remove_ the `descSource` field entirely: if you look at the one place it is used, to pass to `GetDescriptorText`, you'll see that it's not actually needed.)

The existing `NewDefaultEventHandler` function can remain (for backwards compatibility), but can be deprecated. Its `verbose` flag could be used to set verbosity to zero or one.

Finally, you can have your change only print the estimated size when verbosity > 1.

For wording of the message, I would use the term "estimated" instead of "canonical", because I suspect that will be more clear to users that it may not be accurate. (Casual gRPC users may not know what "canonical size" of a protobuf really even means.)

WDYT?

vors:
That sounds like a good plan. I can take a stab at it over the weekend.

vors:
@jhump I updated the diff. I didn't preserver the `NewDefaultEventHandler` signature for backward comparability as you suggested. I'm not sure it's going to be a problem in practice with people using `grpcurl` mostly as a cmd tool. But it's definitely up to you and your know better the level of backward comparability you want to provide with the project. So let me know if you want me to introduce a new `NewDefaultEventHandlerWithVerbosityLevel` helper.

jhump:
> I didn't preserver the `NewDefaultEventHandler` signature for backward comparability as you suggested. I'm not sure it's going to be a problem in practice with people using `grpcurl` mostly as a cmd tool.

Sorry, but this will definitely be a problem. There are numerous consumers of the Go package. This is change is simply *not* worthy of a new major version, which means we need to preserve compatibility for all exported API.

What I was suggesting is to change the usage of the function, in the cmd package, to just use a struct initializer expression and to make the fields exported for that to work. Then you can deprecate that function, instead of making an incompatible change to it. And its implementation will need to translate the `bool` value into a 0 or 1 level.

vors:
Sounds good, let me update the PR.

vors:
Updated the PR. @jhump would appreciate a review when you have a moment!

format.go
```diff
@@ -409,10 +409,13 @@ func RequestParserAndFormatterFor(format Format, descSource DescriptorSource, em
 // safe for use with InvokeRPC as long as NumResponses and Status are not read
 // until the call to InvokeRPC completes.
 type DefaultEventHandler struct {
-	out        io.Writer
-	descSource DescriptorSource
-	formatter  func(proto.Message) (string, error)
-	verbose    bool
+	Out        io.Writer
+	DescSource DescriptorSource
```
jhump, line 413: I mentioned this in an earlier comment, but this field can simply be removed, instead of widening the exported API of this package with an unused field.

The one place it it used can be change to pass `nil` since the function being invoked (`GetDescriptorText`) no longer needs a descriptor source.

format.go
```diff
@@ -409,10 +409,13 @@ func RequestParserAndFormatterFor(format Format, descSource DescriptorSource, em
...
+	Formatter  func(proto.Message) (string, error)
```
jhump, line 414: Let's use `Formatter` as the type here.

format.go
vors: Got it, sorry for catching up so slowly on your comments :)

format.go
vors: :+1:

PR #182, opened 2020-08-18
feat: add support for user-agent header

gszr:
Given value is added to the user-agent header
set by the grpc-go library.

Related: https://github.com/grpc/grpc-go/pull/255

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -37,7 +37,9 @@ import (
 // the response status codes emitted use an offest of 64
 const statusCodeOffset = 64
 
-var version = "dev build <no version set>"
+const no_version = "dev build <no version set>"
+
+var version = no_version
 
 var (
 	exit = os.Exit
@@ -85,6 +87,10 @@ var (
 		is used, this will also be used as the server name when verifying the
 		server's certificate. It defaults to the address that is provided in the
 		positional arguments.`))
+	userAgent = flags.String("user-agent", "", prettify(`
+		If set, the specified value will be added to the User-Agent header set
+		by the grpc-go library.
+		`))
 	data = flags.String("d", "", prettify(`
 		Data for request contents. If the value is '@' then the request contents
 		are read from stdin. For calls that accept a stream of requests, the
@@ -426,6 +432,16 @@ func main() {
 		} else if *authority != "" {
 			opts = append(opts, grpc.WithAuthority(*authority))
 		}
+
+		grpcurlUA := "grpcurl/" + version
+		if version == no_version {
+			grpcurlUA = "grpcurl/dev-build (no version set)"
+		}
+		if *userAgent != "" {
+			grpcurlUA = *userAgent + " " + grpcurlUA
+		}
+		opts = append(opts, grpc.WithUserAgent(grpcurlUA))
+
 		network := "tcp"
 		if isUnixSocket != nil && isUnixSocket() {
 			network = "unix"


cmd/grpcurl/grpcurl.go
```diff
@@ -85,6 +85,10 @@ var (
 		is used, this will also be used as the server name when verifying the
 		server's certificate. It defaults to the address that is provided in the
 		positional arguments.`))
+	user_agent = flags.String("user-agent", "", prettify(`
```
jhump, line 88: The convention for Go vars is camel-case, not snake-case. So `userAgent`.

cmd/grpcurl/grpcurl.go
```diff
@@ -426,6 +430,9 @@ func main() {
 		} else if *authority != "" {
 			opts = append(opts, grpc.WithAuthority(*authority))
 		}
+		if *user_agent != "" {
+			opts = append(opts, grpc.WithUserAgent(*user_agent))
```
jhump, line 434: Oh, neat. While setting user agent, I think it would be good to add grpcurl version info in here, too.

 Maybe something like so:
```go
grpcurlUA := sanitizeUA("grpcurl/" + version)
if version == "dev build <no version set>" {
    grpcurlUA = "grpcurl/dev-build (no version set)"
}
if *userAgent != "" {
    grpcurlUA = sanitizeUA(*userAgent) + " " + grpcurlUA
}
append(opts, grpc.WithUserAgent(grpcurlUA))

// sanitizeUA should replace whitespace chars (space, tab, CR, LF, vertical tab)
// with underscores. Ideally, it would be a little smarter and recognize if the UA
// is already in valid format -- <product>/<version> (<extra>) -- and only replace
// characters that need to be replaced to preserve this.
```

Would you be up for making such a change?

gszr: @jhump I've done a subset of the requested changes. The sanitization, however, I've skipped. Thinking a bit more about it, I feel that enforcing the format would be too restrictive; further, silently modifying the passed user-agent might be confusing. I can see use cases (e.g., testing) where that format is not followed. Thoughts?

cmd/grpcurl/grpcurl.go
```diff
@@ -426,6 +432,16 @@ func main() {
 		} else if *authority != "" {
 			opts = append(opts, grpc.WithAuthority(*authority))
 		}
+		if *userAgent != "" {
```
jhump, line 435: I think you need to remove this outer check, since you check for this inside (and you don't need this flag to short-circuit the `grpcurl/...` part of the UA.

jhump (commented):
Looks good other than one tiny change needed.

cmd/grpcurl/grpcurl.go
gszr: 🤦‍♂️ good catch!

jhump (approved):
Thanks for the contribution!

gszr:
Thanks for reviewing/merging, @jhump! Is there a timeline for the next release? Let me know if there's anything I can help with on it!

gszr:
Hi @jhump - apologies to ask again, but is there a timeline for the next release? I'd be happy to help with it!

jhump:
@gszr, I apologies for the long delays. The notice from GitHub from your last comment got buries in my inbox, so I never even saw it until just now. I have now released v1.8.0, which includes this new flag. Sorry for the long lag between releases.

gszr:
@jhump No worries! Thanks for this great tool and happy new year!

PR #183, opened 2020-08-31
fix latest CI breakages by forking code from grpc-go's interop/testing

jhump:
This will [fix CI](https://travis-ci.com/github/fullstorydev/grpcurl/jobs/379510194). One of the jobs runs _without_ Go modules, explicitly to pull in the latest versions of all dependencies, so we have advance notice of a breakage.

Recently, a breaking change was introduced in the generation of server-side registration code for gRPC services. The repo maintained the previous version of generated code for many packages, to maintain compatibility. But the `interop/grpc_testing` package is not covered by their compatibility guarantees. So our test code that relied on that test package broke.

This repo already had a fork of the proto source. So now we just generate our own copy of `test.pb.go` and use that instead of the unstable package from grpc-go.

While at it, I also moved the `testing` package into `internal`, to make it clear that it is _not_ part of this package's API and is not intended be stable across versions (🤞 hopefully this won't actually break people).

Most of the diff is adding the `internal/` prefix to everywhere that referenced files in the testing folder.


dragonsinth:
> This will [fix CI](https://travis-ci.com/github/fullstorydev/grpcurl/jobs/379510194). One of the jobs runs _without_ Go modules, explicitly to pull in the latest versions of all dependencies, so we have advance notice of a breakage.

Literally without go modules?  Or you mean with no `go.mod` file to use pinned versions?

jhump:
> Literally without go modules? 

It currently does it literally without modules: `GO111MODULE=off`. This is because it is easy to configure a target with different env variables with Travis. Changing it to first delete `go.mod` (but still use modules) would be slightly more involved, like to have the main run command o the `Makefile` behave differently or take that extra step. (To be clear, it wouldn't at all be hard to do it that way -- like one extra line in the `.travis.yaml` and/or the `Makefile`; that's just not the way it's done right now.)

Also worth noting that a lot the setup is legacy and from the process slowly morphig over time. It all started off in "non-module" mode. Then I added `GO111MODULE=on` to make sure things would build with modules and newer versions of Go. And more recently, as everything has shifted over to modules, I intentionally left that one job configured that way so that it could catch issues with latest versions of deps...

PR #188, opened 2020-09-29
format gRPC status errors when using reflection

mkatychev:
Current grpc `status.Status` errors propagated from relection do not respect the `-format-error` flag:

```
$ grpcurl -format-error -H "$EXPIRED_TOKEN" -d '{}' grpc.example.com:8000 method.to.Reflect.Here

Error invoking method "method.to.Reflect.Here": failed to query for service descriptor "method.to.Reflect": rpc error: code = Unauthenticated desc = rpc error: code = Unauthenticated desc = invalid auth token: Token is expired

```

This is due to the error being stringified irrespective of any flags:

https://github.com/fullstorydev/grpcurl/blob/9846afccbc2f34255dfb459dc6f0196a2b6dbe05/invoke.go#L101

---

Propagating the preserved gRPC status error in `gprcurl.InvokeRPC` if `status.FromErrror` is ok correctly produces the expected behaviour when `formatError` is set to true:
```
$ grpcurl -format-error -H "$EXPIRED_TOKEN" -d '{}' grpc.example.com:8000 method.to.Reflect.Here

{
  "code": 16,
  "message": "rpc error: code = Unauthenticated desc = invalid auth token: Token is expired"
}
```

### DIFF
--- a/invoke.go
+++ b/invoke.go
@@ -93,10 +93,15 @@ func InvokeRPC(ctx context.Context, source DescriptorSource, ch grpcdynamic.Chan
 	if svc == "" || mth == "" {
 		return fmt.Errorf("given method name %q is not in expected format: 'service/method' or 'service.method'", methodName)
 	}
+
 	dsc, err := source.FindSymbol(svc)
 	if err != nil {
 		if isNotFoundError(err) {
 			return fmt.Errorf("target server does not expose service %q", svc)
+		}
+		// return the error unstringified if it is a gRPC status.Status error
+		if _, ok := status.FromError(err); ok {
+			return err
 		}
 		return fmt.Errorf("failed to query for service descriptor %q: %v", svc, err)
 	}
--- a/invoke.go
+++ b/invoke.go
@@ -101,7 +101,7 @@ func InvokeRPC(ctx context.Context, source DescriptorSource, ch grpcdynamic.Chan
 		}
 		// return the error unstringified if it is a gRPC status error
 		if statusErr, ok := status.FromError(err); ok {
-			return status.Errorf(statusErr.Code(), "failed to query for service descriptor %q: %v", svc, err)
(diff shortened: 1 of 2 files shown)


invoke.go
```diff
@@ -93,11 +93,16 @@ func InvokeRPC(ctx context.Context, source DescriptorSource, ch grpcdynamic.Chan
...
 		return fmt.Errorf("given method name %q is not in expected format: 'service/method' or 'service.method'", methodName)
 	}
+
 	dsc, err := source.FindSymbol(svc)
 	if err != nil {
 		if isNotFoundError(err) {
 			return fmt.Errorf("target server does not expose service %q", svc)
 		}
+		// return the error unstringified if it is a gRPC status.Status error
+		if _, ok := status.FromError(err); ok {
```
mkatychev, line 103: I realize that this is not a backwards compatible change. I'm not sure how inject any conditional logic short of importing `"cmd/grpcurl".formatError` here, and I don't think that's acceptable.

invoke.go
```diff
@@ -93,11 +93,16 @@ func InvokeRPC(ctx context.Context, source DescriptorSource, ch grpcdynamic.Chan
 	if svc == "" || mth == "" {
 		return fmt.Errorf("given method name %q is not in expected format: 'service/method' or 'service.method'", methodName)
 	}
+
 	dsc, err := source.FindSymbol(svc)
 	if err != nil {
 		if isNotFoundError(err) {
 			return fmt.Errorf("target server does not expose service %q", svc)
 		}
+		// return the error unstringified if it is a gRPC status error
```
jhump, line 102: Seems like this should move up, so you can format "not found" errors, too.

jhump (approved):

jhump:
Thanks for the contribution!

PR #197, opened 2020-11-23
Fix/udpa dependency reference

kai5263499:
Update the upda dependency reference to fix this error I was seeing when running the `go get github.com/fullstorydev/grpcurl/...` installation step

```
package github.com/fullstorydev/grpcurl/cmd/grpcurl
	imports github.com/cncf/udpa/go/udpa/core/v1: cannot find package "github.com/cncf/udpa/go/udpa/core/v1" in any of:
	/usr/local/go/src/github.com/cncf/udpa/go/udpa/core/v1 (from $GOROOT)
	/home/wes/go/src/github.com/cncf/udpa/go/udpa/core/v1 (from $GOPATH)
```

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -13,10 +13,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/fullstorydev/grpcurl"
 	descpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/grpcreflect"
+	"github.com/kai5263499/grpcurl"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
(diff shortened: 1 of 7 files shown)


jhump:
The only changes in here, other than changing a bunch of imports to your own fork (which I clearly will not be able to accept), seems to be the way you are trying to pin the dependency.

However, it should already be pinned, indirectly, by the [grpc-go module](https://github.com/grpc/grpc-go/blob/1f47ba4663831f2a9c28a62a7de3ff8bc45078f0/go.mod#L6).

I think the issue is more complicated and likely has to do with your current working directory when you run `go get ...`. I think the Go tool can end up running in `GOPATH` mode, depending on whether or not you are in a workspace or not. And in `GOPATH` mode, it is not using the `go.mod` constraints.

An issue has already been filed with upstream as this is a breaking change they introduced a little over a week ago: https://github.com/cncf/udpa/issues/42

PR #168, opened 2020-06-23
Bump grpc-go to 1.30.0

erikjoh:
@jhump [Version 1.30.0](https://github.com/grpc/grpc-go/releases/tag/v1.30.0) introduces the new `xds` scheme and some nice improvements to xds in general.

PR #198, opened 2020-11-25
Bump grpc-go to 1.33.1

erikjoh:
Hi @jhump !
I'm looking to use some new xDS bootstrap features that are accessible in versions after 1.30.0 ([this particularity](https://github.com/grpc/grpc-go/pull/3881)).
Looking at the release notes https://github.com/grpc/grpc-go/releases there doesn't seem to be any breaking or noticeable changes since 1.30.0.
This PR is similar to #168 .


erikjoh:
Decided to change this to 1.33.1 because apparently version 1.33.2 and above [uses a deprecated version of github.com/golang/protobuf](https://github.com/grpc/grpc-go/blob/44e408dab41e7c08c1d9369c84a3df5b5c11d886/go.mod#L9) which causes a build failure.

erikjoh:
Closing in favor of #230

PR #205, opened 2021-01-21
add flag for grpc.MaxCallSendMsgSize

binocarlos:
Signed-off-by: Kai Davenport <kaiyadavenport@gmail.com>

Hello! We needed a way to control the maximum size of a *request* message (a.k.a `grpc.MaxCallSendMsgSize`)

The current `-max-msg-sz` flag currently affects `grpc.MaxCallRecvMsgSize`

Changes in this PR:

 * introduce a `-max-req-sz` that, if supplied is assigned to `grpc.MaxCallSendMsgSize`
 * introduce a `-max-res-sz` that, if supplied is assigned to `grpc.MaxCallRecvMsgSize`
 * if `-max-msg-sz` is given - it is applied to both `-max-req-sz` and `-max-res-sz` (providing they are not already defined)

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -125,8 +125,14 @@ var (
 		The maximum total time the operation can take, in seconds. This is
 		useful for preventing batch jobs that use grpcurl from hanging due to
 		slow or bad network links or due to incorrect stream method usage.`))
-	maxMsgSz = flags.Int("max-msg-sz", 0, prettify(`
+	maxReqSz = flags.Int("max-req-sz", 0, prettify(`
+		The maximum encoded size of a request message, in bytes, that grpcurl
+		will accept. If not specified, defaults to 4,194,304 (4 megabytes).`))
+	maxResSz = flags.Int("max-res-sz", 0, prettify(`
 		The maximum encoded size of a response message, in bytes, that grpcurl
+		will accept. If not specified, defaults to 4,194,304 (4 megabytes).`))
+	maxMsgSz = flags.Int("max-msg-sz", 0, prettify(`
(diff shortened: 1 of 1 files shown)


jhump:
@binocarlos, the default "max recv size" is unlimited, so why would you need to specify one?

If it's to avoid sending too large of a message to the server, does that matter? The call will fail either way, whether it's the client preventing the send or the server refusing to accept it.

I'm not sure having the client skip sending is an optimization that is worth broadening the already wide set of command-line options.

binocarlos:
@jhump thanks for your reply 

> the default "max recv size" is unlimited, so why would you need to specify one?

did you mean the default "max send size" here?

the use case is our clients project uses GRPC to send binary uploads to an endpoint - so we need to ensure the **send** size will allow large uploads.

so it's more to allow large uploads than disallow them - the server will be setup to allow the large requests so we just need to ensure the client allows it.

If however - the max *send* size is unlimited by defauilt then we are golden and I don't need this PR :-)

jhump:
@binocarlos, yeah, sorry, I did indeed mean max _send_ size.

Here are the defaults in the gRPC runtime library: https://github.com/grpc/grpc-go/blob/39a500abb9b14b9d9bedb07a8323810a46690c09/clientconn.go#L96

These controls are meant to prevent unbounded memory use. So there's a limit on received messages so that an abusive client can't send an arbitrarily large message and cause the receiver to run out of memory. But for sending, at the point the size is checked, the memory is already consumed because the sender has already constructed the message to send. So it defaults to wide open.

One might want to lower the send limit just to optimize for network utilization: to not bother sending an extremely large message on the wire if we know the server will reject it (due to server having a receive limit configured). But I don't think the network optimization is worth it for grpcurl since it complicates the command-line interface.

Issue #211, opened 2021-02-13
Support for optional on proto3


pcriv:
Proto3 support using the keywork `optional` since 3.12 by adding the flag [--experimental_allow_proto3_optional](https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md#how-to-enable-explicit-presence-in-proto3) to the `protoc` command.

Currently grpcurl fails with this message when:

```
could not parse given files: ./foo.proto:48:5: field BarRequest.foobar: field has label LABEL_OPTIONAL, but proto3 must omit labels other than 'repeated'
```

Is there anyway to tell grpcurl to also use this flag?

bbemis017:
Do we know when this will be released?

jhump:
@bbemis017, here ya go: https://github.com/fullstorydev/grpcurl/releases/tag/v1.8.1

PR #212, opened 2021-02-20
use latest version of protoreflect

jhump:
This also triggers the use of v1.4+ of the protobuf runtime (which begins pulling in packages from the new runtime).

That is why a lot of the non-generated code changed: I changed imports and added `lint:ignore` comments throughout since, in the newer versions of these packages, they are now _deprecated_.

Since the new API is not compatible, it would be non-trivial to migrate to it _and_ would require a v2 and breaking compatibility, since we use the older runtime types in `grpcurl`'s exported API. (If we preserved compatibility, it means we'd be keeping around two copies of many API elements, half of them deprecated. That is its own can of worms since the new API elements would have to have different names, so we'd need two names for everything, and I find just adding "v2" as a suffix to names rather appalling from an aesthetic perspective...)

The majority of changes here are actually to _generated_ code: since this repo is now linking a newer version of the protobuf runtime, I also re-generated all of the `.pb.go` code using the corresponding newer version of `protoc-gen-go`.


jhump:
This will resolve #211 because the latest version of the `protoparse` package (in the `jhump/protoreflect` repo) supports source files that contain "proto3 optional" fields.

internal/testing/jsonpb_test_proto/test_objects.proto
```diff
@@ -1,34 +1,3 @@
-// Go support for Protocol Buffers - Google's data interchange format
```
jhump, line 1: AFAICT, this file did not actually come from Google. I'm not sure how it ended up with this copyright preamble...

dragonsinth:
There's an API?  I thought grpcurl was basically a command line tool.

dragonsinth:
I thought most of the v1->v2 stuff was type aliases anyway...

jhump:
> There's an API? I thought grpcurl was basically a command line tool.

Yes, there's an API. `grpcui` uses as does some other 3rd-party stuff, like uber/prototool.

> I thought most of the v1->v2 stuff was type aliases anyway...

It is. I did not change it because it was necessary for compile. I changed it to get rid of lint/vet errors about using deprecated stuff. The new version deprecates the use of the old import paths.

dragonsinth (approved):

PR #223, opened 2021-04-13
Add option to disable CloseSend()

fiibbb:
The current implementation calls `CloseSend` on bidirectional stream once full request has been read. With `grpc-go`, this results in an `io.EOF` on the server side when server calls `Recv()`. According to http2/grpc semantic, this should result in the stream to be put into half-closed state, where server should no longer try to receive messages from client, but may still choose to keep the stream open. The specific behavior of the server can differ depending on user implementation, so I would like to control whether grpcurl uses `CloseSend()`, depending what server I'm interacting with.

This is not a fully flushed out PR (tests, style, comments etc), but the behavior does match what I want. I'm interested in some early feedback if this is something maintainers would be open to consider.

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -122,6 +122,9 @@ var (
 		probe is sent. If the connection remains idle and no keepalive response
 		is received for this same period then the connection is closed and the
 		operation fails.`))
+	disableHalfClose = flags.Bool("disable-half-close", false, prettify(`
+		If true, the client will not call CloseSend() on the stream after all
+		request messages have been sent.`))
 	maxTime = flags.Float64("max-time", 0, prettify(`
 		The maximum total time the operation can take, in seconds. This is
 		useful for preventing batch jobs that use grpcurl from hanging due to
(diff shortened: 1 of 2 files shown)


jhump:
@fiibbb, I do not understand the value of this.

When a Go server receives `io.EOF` from `Recv()`, that is how it knows the stream is half-closed. The server can continue to process and send messages, and the RPC does not end until server handler returns (or the operation is cancelled/times out).

fiibbb:
> @fiibbb, I do not understand the value of this.
> 
> When a Go server receives `io.EOF` from `Recv()`, that is how it knows the stream is half-closed. The server can continue to process and send messages, and the RPC does not end until server handler returns (or the operation is cancelled/times out).

Some servers do not implement the full "half close" semantics. They may terminate the entire stream when they read an error from `Recv()`. In some cases, this can be considered a bug in the server. In some cases, this is by design (for example, if the responses on the stream are generated by some asynchronous workers, and the server doesn't know when the next response will be available when the stream is established). For these servers, the stream maybe terminated before the server send out any response (since grpcurl caused an `io.EOF` on `Recv` before any response is sent out. To handle these cases, it would be useful to make grpcurl "wait until at least one response arrives".

Issue #224, opened 2021-04-16
add  `grpcurl` ins available via Scoop on windows


funcfoo:
On windows `grpcurl` ins available via Scoop:
```shell
scoop install grpcurl
```

funcfoo:
#225

jhump:
I chose a different route, so I don't have to maintain a registry of all 3rd party packages/installers. See #229

PR #225, opened 2021-04-16
Update README.md

funcfoo:
grpcurl ins available via Scoop

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,11 @@ On macOS, `grpcurl` is available via Homebrew:
 brew install grpcurl
 ```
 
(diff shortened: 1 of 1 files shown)


jhump:
We don't know anything about Scoop nor do we maintain that package. So I'm not too keen on referring to it directly in our README.

However, I see that Scoop is already listed in the repology.org listing for `grpcurl`. So what do you think about this alternative README update: #229? It calls out packages that are maintained by 3rd parties and links to repology.org.

jhump:
I hope my alternative is adequate. If you have any remaining concerns, please let me know via comments on #224.

PR #226, opened 2021-04-18
added condition on logging of JSON Response

0001vrn:
**Rationale:**
In the case of large JSON responses, this will protect us from excessive logging.

More details on : https://github.com/ExpediaGroup/mittens/pull/160

### DIFF
--- a/format.go
+++ b/format.go
@@ -478,7 +478,9 @@ func (h *DefaultEventHandler) OnReceiveResponse(resp proto.Message) {
 	if respStr, err := h.Formatter(resp); err != nil {
 		fmt.Fprintf(h.Out, "Failed to format response message %d: %v\n", h.NumResponses, err)
 	} else {
-		fmt.Fprintln(h.Out, respStr)
+		if h.VerbosityLevel > 0 {
+			fmt.Fprintln(h.Out, respStr)
+		}
 	}
 }
 


jhump (commented):
This is not a backwards compatible change and makes the tool unusable with tools like `jq` since enabling verbose output includes non-JSON log output. A big part of the value of the tool for most users is the response message output, so this is a confusing change to me.

In the description, you mention a concern of very large responses in the output. If that is the concern, I think you just want to pipe the output to `/dev/null` for your case or perhaps use the `-max-msg-sz` flag, to have `grpcurl` refuse to accept messages that are very large.

0001vrn:
Apologies, I shared the original PR which lead me to introduce this change in `grpcurl`. Please have a look and read through. It should give more context and make things clear. 

> This is not a backward compatible change and makes the tool unusable with tools like `jq` since enabling verbose output includes non-JSON log output

Could you help me make this change backward compatible? 
Also, if we enable verbose output, we will still see logs. This change only adds a condition on printing response from gRPC API. 

We’re using `grpcurl` as third party dependency in [mittens](https://github.com/ExpediaGroup/mittens) and it is being used in production with more than 1000 apps running on Kubernetes. And even though we've disabled verbose logging in the [client](https://github.com/ExpediaGroup/mittens/blob/main/pkg/grpc/client.go#L88), it still prints the JSON response from gRPC, so the intention of this change is to just suppress that only. In most cases, gRPC JSON responses are quite large, and this creates excessive logging in the k8s container.

I hope I make sense :)

jhump:
> Could you help me make this change backward compatible?

Not this change, no. But I can certainly help with what you need to accomplish in `mittens` -- it can be done with a fairly trivial change, with no changes at all to `grpcurl`.

The problem is you are currently using `grpcurl.NewDefaultEventHandler`. But the _default_ event handler logs requests and responses to its output. That it's sole purpose!

So you need to create your own implementation of `InvocationEventHandler` instead of using that one. You could always _delegate_ to the default to simplify your implementation:

```go
//  use this to create your handler
delegate := &grpcurl.DefaultEventHandler{
	Out:            os.Stdout,
	Formatter:      formatter,
}
loggingEventHandler := eventHandler{InvocationEventHandler: delegate, logResponses: enableResponseLogging}

// then define eventHandler like so
type eventHandler struct {
	grpcurl.InvocationEventHandler
	logResponses bool
}

// override just the method for logging responses, to do so conditionally
func (h eventHandler) OnReceiveResponse(msg proto.Message) {
	if h.logResponses {
		h.InvocationEventHandler.OnReceiveResponse(msg)
	}
}
```

0001vrn:
I tried the approach which you suggested. I tested it locally, it doesn’t seem to work. 

Could you help me fix it, please? 

https://github.com/0001vrn/mittens/pull/1/files

jhump:
Why are you passing `loggingEventHandler.InvocationEventHandler` to `InvokeRPC`? That's just a default event handler. You want to just pass `loggingEventHandler` -- your custom implementation. As is, it never calls your overridden version of `OnReceiveResponse` but is always calling methods on the underlying default impl.

0001vrn:
Gotcha. That was the missing part. Actually, I got a type issue in golang that's why I passed `loggingEventHandler.InvocationEventHandler` but looks like that type issue is no longer bothering me. 

On that note, I would like to say - Thanks a ton @jhump :D You have been a huge help :)

closing this PR as it is not required now :)

PR #229, opened 2021-04-26
freshen README, point to 3rd party packages for install

### LINKED ISSUE #154: How to install the grpcurl binary on Linux?
I look through this section https://github.com/fullstorydev/grpcurl#binaries, must I build it by myself? The below is the process of the installation, but I don't think it a good way.
```console
# wget https://github.com/fullstorydev/grpcurl/archive/v1.5.1.tar.gz
# tar -zxvf v1.5.1.tar.gz
# cd grpcurl-1.5.1/cmd/grpcurl/
# go build
# ls
go1_10.go  go1_9.go  grpcurl  grpcurl.go  indent_test.go  unix.go
```
So, how to install it by `yum`? Thanks!

jhump:
Resolves #154 and #224.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,32 +49,54 @@ files (containing compiled descriptors, produced by `protoc`) to `grpcurl`.
 
 Download the binary from the [releases](https://github.com/fullstorydev/grpcurl/releases) page.
 
+### Homebrew (macOS)
+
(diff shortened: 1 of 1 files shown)


jameremo (approved):
LGTM. I didn't know about repology.org, TIL.

PR #230, opened 2021-04-28
Bump grpc-go to 1.37.0

erikjoh:
Hi!
We would like to be able to call services using the built in xds resolver:
```
grpcurl ... xds:///myservice ...
```

While this is possible using 1.30.0 ( see #168 ), the newer versions of grpc-go allow you to use [xDS v3 instead of v2](https://github.com/grpc/grpc-go/blob/91d8f0c916d76f2a5aac9e846cd7ffcb838db769/xds/xds.go#L46).
Since the xDS v2 API has been [deprecated](https://www.envoyproxy.io/docs/envoy/latest/api/api_supported_versions) by the Envoy community, it would be nice to have access to v3 in `grpcurl`, so that we eventually can disable the v2 endpoints in the control plane server.
There are of course many other improvements to xDS and grpc-go in this version bump, but this is the primary reason for us to want an update released fairly soon if possible :slightly_smiling_face:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/nishanths/predeclared v0.0.0-20200524104333-86fad755b4d3 // indirect
 	golang.org/x/net v0.0.0-20200625001655-4c5254603344
 	google.golang.org/api v0.29.0 // indirect
-	google.golang.org/grpc v1.30.0
+	google.golang.org/grpc v1.37.0
 	google.golang.org/protobuf v1.25.1-0.20200805231151-a709e31e5d12
 	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
 )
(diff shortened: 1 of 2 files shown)


jhump:
This seems fine, but can you provide any info -- like a PR description -- as to the purpose? Is there something in the new grpc-go runtime that you think grpcurl needs?

erikjoh:
> This seems fine, but can you provide any info -- like a PR description -- as to the purpose? Is there something in the new grpc-go runtime that you think grpcurl needs?

Yes! Got a bit sidetracked waiting for the build to finish but it looks like Travis isn't working at all.
Updated the PR description now.

jhump:
> looks like Travis isn't working at all.

Yeah, sorry about that. Travis CI changed the way they support free builds for OSS software, and we hit a limit so that no builds are triggered anymore.

It's on our TODO list to migrate to Circle CI (along the lines of this similar change in another repo: https://github.com/fullstorydev/hauser/commit/9d1f65ac6c2729839d39bdbf221155f37148747f).

jhump (approved):

menghanl:
Thanks for making the change @erikjoh.

@jhump Can we also cut a release for this? Thanks!

jhump:
@menghanl, https://github.com/fullstorydev/grpcurl/releases/tag/v1.8.1

PR #239, opened 2021-07-09
add cancel-after: cancel without timing out

zellyn:


cmd/grpcurl/grpcurl.go
```diff
@@ -388,6 +394,17 @@ func main() {
 		defer cancel()
 	}
 
+	if *cancelAfter > 0 {
+		timeout := time.Duration(*cancelAfter * float64(time.Second))
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithCancel(ctx)
```
moshegood, line 400: Why not just use `context.WithTimeout(ctx, timeout)`?

jhump:
Hey, @zellyn! How are you doing?

There is already a `max-time` flag. I guess you want it to _cancel_, as opposed to using deadline propagation? Could you instead achieve this _outside_ of grpcurl? Like use `&` to run it in the background and then send the background process a SIGTERM after the timeout? I think that would achieve the same effect, without cluttering the already long list of flags for this thing.

PR #240, opened 2021-07-12
Upgrade golang to 1.15 version

sunil-thorat:
Current golang version v1.13 has some Jfrog XRay security issues. So upgrading to latest v1.16.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,6 @@ matrix:
       env: GO111MODULE=on
     - go: 1.15.x
       env: GO111MODULE=on
-    - go: 1.16.x
-      env: GO111MODULE=on
     - go: tip
       env: GO111MODULE=on
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,10 @@ matrix:
       - VET=1
     - go: 1.14.x
       env: GO111MODULE=on
+    - go: 1.15.x
+      env: GO111MODULE=on
+    - go: 1.16.x
+      env: GO111MODULE=on
     - go: tip
       env: GO111MODULE=on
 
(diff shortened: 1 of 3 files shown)


jhump:
@sunil-thorat, the Go project still supports 1.15. If the issue is not fixed in 1.15, then it sounds like a bug needs to be filed with Go project about whatever security issue exists.

Do you happen to have any links to the issues you're talking about? Are they with particular runtime packages, or with the build tool chain (and modules)?

Is your issue with the distributed binaries? This change alone will not change that: unless something changed in the `go` tool very recently, putting 1.16 in the `go.mod` does not actually _prevent_ building the code with Go 1.15 or earlier. Also note that you can freely build `grpcurl` yourself, using Go 1.16, without needing any changes to this repo.

sunil-thorat:
Thanks @jhump. The security issues are mainly in golang 1.13 runtime with which distributed binaries are built. They are already raised on `Golang` projects and most of them are fixed in latest 1.15, 1.16 versions. Below is list of issues. 
image

Again we have already started using grpcurl build with Go 1.16 in which we're not seeing above issues.

jhump:
> The security issues are mainly in golang 1.13 runtime with which distributed binaries are built.

As I mentioned in my comment, this line in `go.mod` does not dictate what version the binaries are built with. In fact, the binaries published to GitHub for the latest release (v1.8.1) were actually built with Go 1.15.5. However, it does look like the Docker build is using 1.13 (so updating the `Dockerfile` is the right way to fix that).

If you're willing to push a little further on this PR, let's use `1.15` in the `go.mod` file and then also update that `Dockerfile` to use 1.15, too.

jhump:
Oops, sorry, didn't mean to close this... just meant to leave a comment.

jhump (approved):
LGTM

PR #241, opened 2021-07-12
allow callers of BlockingDial to override grpc.FailOnNonTempDialError

jhump:
If a caller of `grpc.BlockingDial` wanted the dial operation to keep retrying in the face of non-temporary errors (like "connection refused"), it is not currently possible. Even if they tried to use an explicit `grpc.FailOnNonTempDialError(false)` option, it is silently overridden by the `BlockingDial` function.

This PR fixes that. So the caller can provide an explicit option and override the default behavior of `BlockingDial` with regards to failing fast. This will be used to remedy https://github.com/fullstorydev/grpcui/issues/101.


jameremo (approved):
LGTM!

PR #242, opened 2021-07-12
use newer protoreflect

jhump:
Here are the release notes: https://github.com/jhump/protoreflect/releases/tag/v1.9.0

This brings `grpcurl`'s parsing of proto sources closer to matching the behavior of protoc. It also adds support for custom options for oneof elements in source files (which was previously broken -- so source files that had custom oneof options might not be parseable by grpcurl).


jameremo (approved):
LGTM!

PR #243, opened 2021-07-16
Tidy up dependencies

VeryStrongFingers:
This PR is simply the result of running `go mod tidy`, nothing more.

Mainly interested in cleaning up the transitive dependency to `github.com/gogo/protobuf` because of CVE-2021-3121
(according to go mod tidy you don't even need the module at all)


jhump:
> Mainly interested in cleaning up the transitive dependency to `github.com/gogo/protobuf`

Gogo does not appear in the transitive dependencies. It's not in `go.mod` and `go mod why ...` says it's not here:

```
> go mod why github.com/gogo/protobuf
go: downloading github.com/gogo/protobuf v1.2.1
# github.com/gogo/protobuf
(main module does not need package github.com/gogo/protobuf)
```

Are you only referring to the reference to Gogo in the `go.sum`? The `go.sum` file is _not_ a list of dependencies. It is just a checksum database, for security purposes.

go.mod
```diff
@@ -3,14 +3,12 @@ module github.com/fullstorydev/grpcurl
...
 
 require (
+	cloud.google.com/go v0.56.0 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/golang/protobuf v1.4.2
-	github.com/gordonklaus/ineffassign v0.0.0-20200309095847-7953dde2c7bf // indirect
-	github.com/goreleaser/goreleaser v0.134.0 // indirect
 	github.com/jhump/protoreflect v1.9.0
-	github.com/nishanths/predeclared v0.0.0-20200524104333-86fad755b4d3 // indirect
-	golang.org/x/net v0.0.0-20200625001655-4c5254603344
```
jhump, line 11: This is the only real dependency removed. I think it highly unlikely that this referred to Gogo, even indirectly.

The other dependencies were indirect because they are referenced in the `Makefile`, but not by any of the code. So they cannot be removed -- otherwise, the tools aren't correctly pinned to these versions.

VeryStrongFingers: Some security scanners consider go.sum entries to be "transitive dependencies" I guess because they're contextually more than just checksums, their presence indicates that some nested dependency has their go.mod set to a version in the lockfile, so in that sense it is transitive. 
There's even reference to go.sum being for transitive deps it on the Go blog: https://blog.golang.org/migrating-to-go-modules
> go mod tidy finds all the packages transitively imported by packages in your module. ...

Regardless of semantics I see what you mean about the Makefile, it's definitely a bit awkward to `go mod tidy` with that in mind - any ideas on how you'd like to move forward? or shall we just close it for now

jhump: So I played around with adding a `tools.go` file -- a pattern I've seen for pinning versions of tools that are not actually imported by code. Here's the file:
```go
// +build tools

package grpcurl

import (
	"github.com/gordonklaus/ineffassign"
	"github.com/goreleaser/goreleaser"
	"github.com/kisielk/errcheck"
	"github.com/nishanths/predeclared"
	"honnef.co/go/tools/cmd/staticcheck"
)
```
If you create this file and then run `go mod tidy`, it will not remove the reference to gogo.

However, I think it's pretty clear that `go.sum` is _not_ meant to be the source of all of your transitive deps, just based on the fact that a Go module can only use _exactly one_ major version of any dependency. However, you'll see that the `go.sum` file has _many_ versions for the same dependencies, even after running `go mod tidy`.

What's happening is that `go.sum` includes all transitive deps, including the full transitive closure _requested_ by dependencies -- not just the resolved versions. So dependency A might want foo@v1.1 and my module or some other dependency wants foo@v1.4. In this case, v1.1 _and its transitive dependencies_ (which might have included gogo) will be checksum'ed and written into `go.sum`, even though v1.4 and its transitive dependencies are the ones actually used (which may no longer incude gogo).

Unfortunately the `go mod ...` tools are not powerful enough to tell me how gogo got into the `go.sum` file. But, even with the `tools.go` file present, `go mod why github.com/gogo/protobuf` still reports that gogo is not used.

jhump (commented):
If you reset back to master, add the `tools.go` file (see other comment), and then re-run `go mod tidy`, you'll see the changes are much smaller.

If you want to do that and push to this PR, I would probably accept and merge that. However, as I mentioned in the comment, that will not eliminate gogo from the `go.sum` file.

VeryStrongFingers:
Thanks for keeping me in the loop @jhump!

PR #245, opened 2021-07-27
Support SSLKEYLOGFILE environment variable for secrets logging

igorwwwwwwwwwwwwwwwwwwww:
I was trying to debug an issue with wireshark via:

```
$ grpcurl -vv -cacert gdk/localhost.crt localhost:1337 list
Failed to list services: rpc error: code = Unavailable desc = Service Unavailable: HTTP status code 503; transport: received the unexpected content-type "text/html"
```

Wireshark supports [consuming a key log file](https://wiki.wireshark.org/TLS#Using_the_.28Pre.29-Master-Secret), which is supported by several clients using `SSLKEYLOGFILE`, including:

* [Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format)
* Chrome
* [Curl](https://everything.curl.dev/usingcurl/tls/sslkeylogfile)

This patch adds the same capability to grpcurl.

I tested with wireshark and am able to peek inside the TLS packets.

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -408,11 +408,20 @@ func main() {
 		}
 		var creds credentials.TransportCredentials
 		if !*plaintext {
-			var err error
-			creds, err = grpcurl.ClientTransportCredentials(*insecure, *cacert, *cert, *key)
+			tlsConf, err := grpcurl.ClientTLSConfig(*insecure, *cacert, *cert, *key)
 			if err != nil {
-				fail(err, "Failed to configure transport credentials")
+				fail(err, "Failed to configure TLS config")
 			}
+
+			if os.Getenv("SSLKEYLOGFILE") != "" {
+				w, err := os.OpenFile(os.Getenv("SSLKEYLOGFILE"), os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+				if err != nil {
+					fail(err, "Could not open SSLKEYLOGFILE: %v")
+				}
+				tlsConf.KeyLogWriter = w
+			}
+
+			creds := credentials.NewTLS(tlsConf)
 
 			// can use either -servername or -authority; but not both
 			if *serverName != "" && *authority != "" {
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -410,13 +410,14 @@ func main() {
 		if !*plaintext {
 			tlsConf, err := grpcurl.ClientTLSConfig(*insecure, *cacert, *cert, *key)
 			if err != nil {
-				fail(err, "Failed to configure TLS config")
+				fail(err, "Failed to create TLS config")
 			}
 
-			if os.Getenv("SSLKEYLOGFILE") != "" {
-				w, err := os.OpenFile(os.Getenv("SSLKEYLOGFILE"), os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+			sslKeylogFile := os.Getenv("SSLKEYLOGFILE")
+			if sslKeylogFile != "" {
+				w, err := os.OpenFile(sslKeylogFile, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
 				if err != nil {
-					fail(err, "Could not open SSLKEYLOGFILE: %v")
+					fail(err, "Could not open SSLKEYLOGFILE %s", sslKeylogFile)
 				}
 				tlsConf.KeyLogWriter = w
 			}
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -508,9 +508,11 @@ func makeTemplate(md *desc.MessageDescriptor, path []*desc.MessageDescriptor) pr
 	return dm
 }
 
-// ClientTransportCredentials is a thin wrapper around ClientTLSConfig, kept for BC.
-// ClientTLSConfig offers more flexibility, as the caller can customize the tls.Config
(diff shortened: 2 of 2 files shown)


grpcurl.go
```diff
@@ -542,6 +542,14 @@ func ClientTransportCredentials(insecureSkipVerify bool, cacertFile, clientCertF
 		tlsConf.RootCAs = certPool
 	}
 
+	if os.Getenv("SSLKEYLOGFILE") != "" {
```
jhump, line 545: I don't think this belongs in this helper function -- this seems too subtle for a library function, especially with no way to programmatically ignore the env var. And we can't add this flag as a new parameter since that would not be a backwards-compatible change.

Instead, I think we should add a new exported function `ClientTLSConfig` that is most of this function, but just returns `&tlsConf`. This existing function can be re-implemented to delegate to the new one. We could then move most of the Go doc to the new function and have this function's doc that it is a convenience shortcut that calls `ClientTLSConfig` followed by `credentials.NewTLS(result)`.

Once that refactor is done, the caller of this (`grpcurl`'s main method) can directly call `ClientTLSConfig` and there is where we should examine the environment variable and set the `KeyLogWriter` field, before calling `credentials.NewTLS`.

Does that all make sense?

igorwwwwwwwwwwwwwwwwwwww:
Testing it out:

```
$ docker build -t grpcurl .

$ docker run -it -e SSLKEYLOGFILE=/sslkeylog.txt grpcurl 127.0.0.1:1234 list

Could not open SSLKEYLOGFILE: open /sslkeylog.txt: permission denied

$ docker run -it -v $(pwd)/tmp:/tmp -e SSLKEYLOGFILE=/tmp/sslkeylog.txt --user root grpcurl 127.0.0.1:1234 list

Failed to dial target host "127.0.0.1:1234": dial tcp 127.0.0.1:1234: connect: connection refused

$ ls -lah tmp/sslkeylog.txt
```

grpcurl.go
```diff
@@ -508,11 +508,23 @@ func makeTemplate(md *desc.MessageDescriptor, path []*desc.MessageDescriptor) pr
...
+func ClientTransportCredentials(insecureSkipVerify bool, cacertFile, clientCertFile, clientKeyFile string) (credentials.TransportCredentials, error) {
+	tlsConf, err := ClientTLSConfig(insecureSkipVerify, cacertFile, clientCertFile, clientKeyFile)
+	if err != nil {
+		return nil, err
+	}
+
+	return credentials.NewTLS(tlsConf), nil
+}
+
+// ClientTLSConfig builds transport credentials for a gRPC client using the
```
jhump, line 523:
```suggestion
// ClientTLSConfig builds transport-layer config for a gRPC client using the
```

grpcurl.go
```diff
@@ -508,11 +508,23 @@ func makeTemplate(md *desc.MessageDescriptor, path []*desc.MessageDescriptor) pr
 	return dm
 }
 
-// ClientTransportCredentials builds transport credentials for a gRPC client using the
+// ClientTransportCredentials is a thin wrapper around ClientTLSConfig, kept for BC.
+// ClientTLSConfig offers more flexibility, as the caller can customize the tls.Config
+// struct.
```
jhump, line 513:
```suggestion
// ClientTransportCredentials is a helper function that constructs a TLS config with
// the given properties (see ClientTLSConfig) and then constructs and returns gRPC
// transport credentials using that config.
//
// Deprecated: Use grpcurl.ClientTLSConfig and credentials.NewTLS instead.
```

cmd/grpcurl/grpcurl.go
```diff
@@ -408,12 +408,21 @@ func main() {
 		}
 		var creds credentials.TransportCredentials
 		if !*plaintext {
-			var err error
-			creds, err = grpcurl.ClientTransportCredentials(*insecure, *cacert, *cert, *key)
+			tlsConf, err := grpcurl.ClientTLSConfig(*insecure, *cacert, *cert, *key)
 			if err != nil {
-				fail(err, "Failed to configure transport credentials")
+				fail(err, "Failed to configure TLS config")
```
jhump, line 413:
```suggestion
				fail(err, "Failed to create TLS config")
```

cmd/grpcurl/grpcurl.go
```diff
@@ -408,12 +408,21 @@ func main() {
...
 			}
 
+			if os.Getenv("SSLKEYLOGFILE") != "" {
+				w, err := os.OpenFile(os.Getenv("SSLKEYLOGFILE"), os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+				if err != nil {
+					fail(err, "Could not open SSLKEYLOGFILE: %v")
```
jhump, line 419: Let's factor out the file name into a variable. Also, the call to `fail` has a `%v` in the message but no corresponding argument.
```suggestion
			sslKeylogFile := os.Getenv("SSLKEYLOGFILE")
			if sslKeylogFile != "" {
				w, err := os.OpenFile(sslKeylogFile, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
				if err != nil {
					fail(err, "Could not open SSLKEYLOGFILE %s", sslKeylogFile)
```

jhump (commented):
Thanks for making the changes! A left a few final suggestions -- mostly nits related to wording in comments/messages.

igorwwwwwwwwwwwwwwwwwwww:
Thanks, applied!

jhump (approved):

Issue #234, opened 2021-06-08
ARM Support on Docker Image


pquerna:
It looks like goreleaser is producing arm64 binaries, eg on the 1.8.1 release:
https://github.com/fullstorydev/grpcurl/releases/tag/v1.8.1

But the current Dockerfile is only published with amd64 support to dockerhub: 
https://hub.docker.com/r/fullstorydev/grpcurl/tags?page=1&ordering=last_updated

jhump:
@pquerna, a PR was just merged to do multiple architectures for docker. I haven't built and pushed anything for ARM yet. But I don't have access to ARM hardware to test. Could I ping you on this issue when I do, and you can pull and verify the image for me?

jhump:
The last few releases have included arm64 images.

PR #251, opened 2021-08-30
Building multiarch docker image: amd, s390x

namrata-ibm:
This PR aims at building and pushing multi-arch image on docker hub.
New architectures can be easily added to the platforms to generate manifests for others.

### DIFF
--- a/releasing/do-release.sh
+++ b/releasing/do-release.sh
@@ -41,13 +41,15 @@ $PREFIX git checkout go.mod go.sum
 $PREFIX docker login
 echo "$VERSION" > VERSION
 
-# Enable qemu and binfmt support
+# Docker Buildx support is included in Docker 19.03
+# Below step installs emulators for different architectures on the host
+# This enables running and building containers for below architectures mentioned using --platforms
 $PREFIX docker run --privileged --rm tonistiigi/binfmt:qemu-v6.1.0 --install all
 # Create a new builder instance
 export DOCKER_CLI_EXPERIMENTAL=enabled
 $PREFIX docker buildx create --use --name multiarch-builder
 # push to docker hub, both the given version as a tag and for "latest" tag
-$PREFIX docker buildx build --platform linux/amd64,linux/s390x --tag fullstorydev/grpcurl:${VERSION} --tag fullstorydev/grpcurl:latest --push --progress plain --no-cache .
+$PREFIX docker buildx build --platform linux/amd64,linux/s390x,linux/arm64 --tag fullstorydev/grpcurl:${VERSION} --tag fullstorydev/grpcurl:latest --push --progress plain --no-cache .
 rm VERSION
 
 # Homebrew release
--- a/releasing/do-release.sh
+++ b/releasing/do-release.sh
@@ -40,12 +40,15 @@ $PREFIX git checkout go.mod go.sum
 # if there are no valid current credentials.
 $PREFIX docker login
 echo "$VERSION" > VERSION
-$PREFIX docker build -t "fullstorydev/grpcurl:${VERSION}" .
-rm VERSION
+
+# Enable qemu and binfmt support
+$PREFIX docker run --privileged --rm tonistiigi/binfmt:qemu-v6.1.0 --install all
+# Create a new builder instance
+export DOCKER_CLI_EXPERIMENTAL=enabled
+$PREFIX docker buildx create --use --name multiarch-builder
 # push to docker hub, both the given version as a tag and for "latest" tag
-$PREFIX docker push "fullstorydev/grpcurl:${VERSION}"
-$PREFIX docker tag "fullstorydev/grpcurl:${VERSION}" fullstorydev/grpcurl:latest
-$PREFIX docker push fullstorydev/grpcurl:latest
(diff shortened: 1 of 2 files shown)


namrata-ibm:
@jhump could you please check and let me know your comments? Thank you!

namrata-ibm:
@jhump Could you please have a look?

jhump:
I like the idea. This could also be used to resolve #234.

However, I have no access to hardware for these platforms, so I do not know how I could test the images. That makes me nervous in supporting them.

I know the same argument could be made of the compiled binaries in the GitHub release artifacts, which I _have_ included (even for platforms on which I can't test). I guess I trust Go cross-compilation results, having used them a lot. However, I have never used multi-arch support in Docker, and it's even included as "experimental feature". So I guess I'm not as trusting.

Do you have a proposal for how I could verify these images?

namrata-ibm:
@jhump Thank you for your feedback.

We do understand your concern, however, many repositories have enabled buildx for generating multi-arch images now(some of them enabled by our team) - [kubernetes ](https://github.com/kubernetes/kubernetes/blob/master/test/images/README.md#prerequisites), [jaeger](https://github.com/jaegertracing/jaeger/blob/e345aa76f24cf3e4eb1a5b0ecbf378735e56ac5c/CHANGELOG.md) , [kubernetes-csi](https://github.com/kubernetes-csi) amongst the few I can quickly list here. 

Regarding the hardware access to test images, we can provide s390x VMs to you to validate the images at your end. 
Please let me know if it helps.

releasing/do-release.sh
```diff
@@ -40,12 +40,15 @@ $PREFIX git checkout go.mod go.sum
 # if there are no valid current credentials.
 $PREFIX docker login
 echo "$VERSION" > VERSION
-$PREFIX docker build -t "fullstorydev/grpcurl:${VERSION}" .
-rm VERSION
+
+# Enable qemu and binfmt support
+$PREFIX docker run --privileged --rm tonistiigi/binfmt:qemu-v6.1.0 --install all
```
jhump, line 45: What is this for? Does this install something on the host that is needed for the `buildx` stuff below? If so, maybe elaborate in the comment above.

releasing/do-release.sh
```diff
@@ -40,12 +40,15 @@ $PREFIX git checkout go.mod go.sum
...
+# Create a new builder instance
+export DOCKER_CLI_EXPERIMENTAL=enabled
+$PREFIX docker buildx create --use --name multiarch-builder
 # push to docker hub, both the given version as a tag and for "latest" tag
-$PREFIX docker push "fullstorydev/grpcurl:${VERSION}"
-$PREFIX docker tag "fullstorydev/grpcurl:${VERSION}" fullstorydev/grpcurl:latest
-$PREFIX docker push fullstorydev/grpcurl:latest
+$PREFIX docker buildx build --platform linux/amd64,linux/s390x --tag fullstorydev/grpcurl:${VERSION} --tag fullstorydev/grpcurl:latest --push --progress plain --no-cache .
```
jhump, line 50: Let's add `linux/arm64`, too, to resolve #234.

BTW, I don't have any experience using multi-platform docker images. Since this statement uses tags like `latest` and `${VERSION}`, I guess the platform does not need to be part of the tag, and that Docker hub (or other image registries) has a separate way of partitioning images for different platforms?

releasing/do-release.sh
namrata-ibm: Updated the comment, Also had missed earlier to mention about buildx added in docker 19.03 onward. Updated the same.

releasing/do-release.sh
namrata-ibm: > Let's add linux/arm64, too, to resolve #234.

I have added arm64, it should work, however please note that **I do not have access to verify image on arm.** Someone using arm can help us validate. I have verified the images on amd and s390x so far.
Also please note that 'go build' step takes higher time for arm and s390x than amd as of now. Hopefully it will improve in future releases.

> BTW, I don't have any experience using multi-platform docker images. Since this statement uses tags like latest and ${VERSION}, I guess the platform does not need to be part of the tag, and that Docker hub (or other image registries) has a separate way of partitioning images for different platforms?

That's right, docker uses manifest lists for a set of platforms and pulls the correct image based on where it is running. [This ](https://www.docker.com/blog/docker-official-images-now-multi-platform/)describes it well.
I have kept the same image names and tags (latest and $VERSION), this will ensure existing amd users are not affected!

jhump (approved):

releasing/do-release.sh
jhump: When I build and push an arm release, I will see if @pquerna can test an ARM image when I build and push one (that's who requested ARM support in docker).

namrata-ibm:
Thanks @jhump for your support!

jhump:
Ugh, another botched release. It looks like I simply _cannot_ build this for s390x. Apparently it's an issue in qemu on OS X. The work-around I've seen is to use `-cpu qemu,vx=off`, however I cannot find any documentation in docker or binfmt as to how I might supply that command-line arg to qemu during s390x builds.

So I'm afraid I am _reverting_ part of this pull request, removing support for s390x for the docker images that get pushed. The next release will still have published binaries for s390x. But if you want official docker images, I'll need more help in a workable solution.

PR #260, opened 2021-10-07
Spurred by depend-a-bot warnings: dependency hygiene

jhump:
Sadly, this repo does not actually depend on the two libraries that depend-a-bot warns about. The tool is not smart enough to examine the _actual_ dependency graph and instead considers everything in `go.sum`, which is stupid.

These are the libraries that depend-a-bot is complaining about as having security issues:
1. `github.com/ulikunitz/xz` (versions prior to v0.58; `go.sum` refers to two different versions)
2. `github.com/dgrijalva/jwt-go` (all versions)

Anyhow, now most of the CI checks are using Go 1.16, we can use a new feature in Go 1.16: using `go install ...@version` to pin versions of tools, instead of pinning them in `go.mod`.

So I ran `go mod tidy`, which removed all of the indirect dependencies (those tools) and updated the `Makefile` to explicitly indicate a version there instead. Finally, I ran `make ci`, to make sure `go.sum` was fully re-populated.

This removes the two above libraries from `go.sum`. I'll next apply this same change to grpcui and grpchan.


jhump:
> will this locally dirty the go.mod file whenever you run it?

Nope. New feature in Go 1.16: https://golang.org/doc/go1.16#go-command

PR #261, opened 2021-10-07
Print to stdout when the client receives an OK code

tomoyamachi:
Currently, if grpcurl receives an OK code, it will exit without outputting anything.
This PR improve the output: the result will be output to stdout when got an OK code.


tomoyamachi:
Okay. I will check with the exit code.
Thank you for your reply.

PR #263, opened 2021-10-07
Set value of creds instead of creating new variable

jeffDuce:
The `creds` variable was being redeclared instead of having the existing variable's value set.


jeffDuce:
> Does this mean that TLS was just completely broken?

It was for me. Should be fine now though.

PR #267, opened 2021-11-03
feat: Add dependabot configuration

wwuck:
Add dependabot configuration to get automated version and security
updates on go.mod and GitHub Actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    # Check for updates once a day
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    # Check for updates once a week
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,6 @@ version: 2
 updates:
   - package-ecosystem: "gomod"
     directory: "/"
-    # Check for updates once a day
+    # Check for updates once a week
     schedule:
-      interval: "daily"
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,8 +5,3 @@ updates:
     # Check for updates once a day
     schedule:
       interval: "daily"
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    # Check for updates once a week
-    schedule:
-      interval: "weekly"


jeffwidman:
Nudge @jhump it'd be great to see this land.

jhump:
I'm not too familiar with this. Is the idea that it checks every day or every week (via a GitHub action), and then automatically opens a PR to update? I'm trying to get a feel for high the extra maintenance burden might be, to be reviewing and reacting to suggested updates.

jeffwidman:
It's a version update checker provided by Github. You can configure the frequency of checks. In this cae, @wwuck configured it to check daily for updates to go modules, and to check weekly for updates to the github actions that are used by this repo.'

More details: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates

As someone who maintains multiple open source projects, it's been very helpful on the maintenance side because it keeps my deps up to date automatically, but more importantly it pulls in the changelog, commit lists, etc into the PR description so that I can quickly review what changed and evaluate the risk of pulling in the dep bump. Otherwise I have to manually go track all that down and just takes more time.

jhump:
> As someone who maintains multiple open source projects, it's been very helpful on the maintenance

If it was a priority to keep all deps up-to-date, I could see this being useful. Not that we don't want to keep things up-to-date, but it's just not been a priority. So my concern is how often it will be sending PRs. If it's very often, then this will be too much overhead.

Do you have any anecdotal experience about how many PRs per week a single repo can generate? I know it depends on the actual dependencies, their number and how often those dependencies are updated/get new versions published. But I'm looking for an informed guess so I can then provide feedback on this PR and what is the right cadence to start with.

jeffwidman:
Given the small number of deps listed in https://github.com/fullstorydev/grpcurl/blob/master/go.mod, and the fact that dependabot currently _only_ bumps direct deps, not indirect deps, then I suspect these deps only get updated a couple of times per month. 

If you're worried about too much work, then change the dependabot check frequency for go modules to weekly or monthly. Personally I prefer to merge more often so I ahve smaller diffs so that I get a unit test run against both versions, plus if something breaks, I can `git bisect` to quickly track down exactly what upstream change broke, but up to you.

It also helps that merging the dependabot PR's takes about 10 seconds and two clicks to approve, then merge. The main 
thing that consumes time is reviewing the upstream commit changelog, but merging less frequently doesn't really solve that... ie, if you checked monthly  and someone released two versions, you now have two version changelogs to check.

Personally, I don't really care, however you want to run your ship. My main request is that deps aren't 6+ months out of date, esp if there's a CVE etc. I'd probably suggest weekly, that way if someone pushes a version and then retracts it almost immediately due to a bug you're less likely to pickup the bleeding edge.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+  - package-ecosystem: "gomod"
+    directory: "/"
+    # Check for updates once a day
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    # Check for updates once a week
+    schedule:
+      interval: "weekly"
```
jhump, line 12: This whole section is useless, right? There are no workflows or actions defined in this repo, so I don't think there is anything to check here.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    # Check for updates once a day
```
jhump, line 5: Can we do once every 2 weeks?

.github/dependabot.yml
wwuck: Yes, I think I just copied from an existing repository so I will remove it here.

.github/dependabot.yml
wwuck: Unfortunately dependabot doesn't support that.  Would you prefer weekly or monthly?

https://github.com/dependabot/dependabot-core/issues/2038

jhump: Let's do weekly.

wwuck:
@jhump There is also an option to get security updates via dependabot Pull Requests, and from reading the documentation I think this is enabled by default for supported GitHub repositories unless it has been explicitly disabled.

https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/about-dependabot-security-updates

https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/configuring-dependabot-security-updates

jhump (approved):

jeffwidman:
Thanks @jhump!

PR #281, opened 2022-01-25
document how to pass payload from file, fixes #279

### LINKED ISSUE #279: add/document way of passing inputs from a file
Hello!

I'm trying out grpcurl to test a grpc endpoint. I'm having a hard time doing so cause grcpurl does not support passing a input file. 

Please add a way or document (if it already exists) how to load input data from a file. 

If I pass a large json either directly with `-d "myjson"` or with the `-d @ .. <<EOM ` way, I always get "too many arguments" as an error.

domdorn:


README.md
```diff
@@ -146,6 +146,12 @@ grpcurl -d @ grpc.server.com:443 my.custom.server.Service/Method <<EOM
 EOM
 ```
 
+Sometimes it might be necessary to read the payload from a file. This worked for some users:
+
+```shell
+grpcurl -d "$(cat my.json)" grpc.server.com:443 my.custom.server.Service/Method
```
jhump, line 152: I think the more straight-forward way, that doesn't require this particular feature in your shell, is to use the `@` value as the input, very much like the example above.
```suggestion
grpcurl -d @ grpc.server.com:443 my.custom.server.Service/Method < my.json
```

README.md
```diff
@@ -146,6 +146,12 @@ grpcurl -d @ grpc.server.com:443 my.custom.server.Service/Method <<EOM
 EOM
 ```
 
+Sometimes it might be necessary to read the payload from a file. This worked for some users:
```
jhump, line 149: "This worked for some users" -- this wording feels really non-committal. We can make it authoritative.
```suggestion
To use a file as the request payload, redirect the file to `grpcurl`'s stdin:
```

PR #286, opened 2022-02-05
Bump grpc lib and fix deprecations

jeffwidman:
I took #283 and fixed the deprecations.


cmd/grpcurl/grpcurl.go
```diff
@@ -438,9 +438,7 @@ func main() {
 			}
 
 			if overrideName != "" {
-				if err := creds.OverrideServerName(overrideName); err != nil {
-					fail(err, "Failed to override server name as %q", overrideName)
-				}
+				opts = append(opts, grpc.WithAuthority(overrideName))
```
jeffwidman, line 441: I'd appreciate a second set of eyes on this... as best I can tell this is the way to do it now, here's the relevant warning:
```
cmd/grpcurl/grpcurl.go:441:15: creds.OverrideServerName is deprecated: use grpc.WithAuthority instead. Will be supported throughout 1.x.  (SA1019)
```

jhump: This looks right to me.

It's interesting that they are deprecating this function. I had intentionally added the above checks, to disallow specifying _different_ `-servername` and `-authority` values, because it seemed a bit dubious (and possibly strange/bad from a security perspective). But if they are getting rid of this override name, then it won't even be possible to configure the client with different values for these two. So I guess my addition of the check above was prescient...

jhump (approved):
LGTM

cmd/grpcurl/grpcurl.go
jeffwidman: I don't have a lot of context on these overrides... Does this mean `servername` should be removed altogether as an override option? (Or at least hidden from the help menu and marked deprecated in the code so that it's not backwards breaking.)

PR #306, opened 2022-05-03
feat: add support for HTTPS_PROXY

### LINKED ISSUE #166: Option to add proxy
Is there any option to forward the traffic via any app proxy? This will help in easy edit of json requests and understanding of full http requests.

denysvitali:
This commit adds support for the `HTTPS_PROXY` environment variable.
At the moment the `NO_PROXY` environment variable is ignored, please
be aware of that!

This should close #166

On top of that, this adds support for `SOCKS_PROXY` via the [ProxyFromEnvironment](https://pkg.go.dev/golang.org/x/net/proxy#FromEnvironment) method of [`golang.org/x/net/proxy`](https://pkg.go.dev/golang.org/x/net/proxy)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,8 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/golang/protobuf v1.5.2
 	github.com/jhump/protoreflect v1.10.3
+	github.com/mwitkow/go-http-dialer v0.0.0-20161116154839-378f744fb2b8 // indirect
+	golang.org/x/net v0.0.0-20200822124328-c89045814202
 	golang.org/x/text v0.3.7 // indirect
 	google.golang.org/grpc v1.44.0
 	google.golang.org/protobuf v1.27.1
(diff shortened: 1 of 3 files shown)


dragonsinth:
Hi @denysvitali thanks for the contribution.  Is there a reason we'd need a new third party lib instead of using Go's native proxying?

CpuID:
can confirm this is working great for us - would love to see it merged 👍

denysvitali:
> Hi @denysvitali thanks for the contribution. Is there a reason we'd need a new third party lib instead of using Go's native proxying?

https://github.com/mwitkow/go-http-dialer README explains it:

> Some enterprises have fairly restrictive networking environments. They typically operate [HTTP forward proxies](https://en.wikipedia.org/wiki/Proxy_server) that require user authentication. These proxies usually allow HTTPS (TCP to :443) to pass through the proxy using the [CONNECT](https://tools.ietf.org/html/rfc2616#section-9.9) method. The CONNECT method is basically a HTTP-negotiated "end-to-end" TCP stream... which is exactly what [net.Conn](https://golang.org/pkg/net/#Conn) is :)

[net/proxy](https://pkg.go.dev/golang.org/x/net/proxy) cannot be used here AFAIK because you can't have a dialer that uses TCP over HTTP. The only way to do it is to use a SOCKS5 proxy.

dragonsinth:
I think I'd be trivially fine with a patch to just change  `(&net.Dialer{}).DialContext(ctx, network, address)` => `proxy.Dial(ctx, network, address)` which would solve at least some of the cases.  But the http dialer stuff seems... very unusual and special case.

L11R:
Release with this patch: https://github.com/L11R/grpcurl/releases/tag/v1.8.8

vpopescu:
Hi, what exactly is the status of this patch?
I was using the 1.9.1, and could not get the http_proxy to work.
However downloading the version mentioned above (1.8.8) does work.

So was this patch not carried forward, or did it somehow change and require different configuration?

PR #309, opened 2022-05-10
new feature: use grpcurl as an offline tool to convert JSON into proto

irsl:
The tool then can be used like this:

```
$ ./grpcurl --import-path  /data/app-engine-experiments/2022/protos -proto remote_api.proto -d '{"service_name":"sdf","method":"method","request":"foo"}' -convert-message remote_api.Request | protoc --decode_raw
2: "sdf"
3: "method"
4: "foo"
```


dragonsinth:
This seems kind of out of scope for grpcurl.

PR #311, opened 2022-05-24
build alpine base image

BenRedmondBenham:
Could you push an alpine base image version?

Then you could support more complex commands like this:

```
grpcurl -plaintext \
             -H "Authorization: Bearer $$(cat /var/run/secrets/tokens/vault-token)" \
             localhost:80 \
             example.v1.Service/Func
```

This works with a common auth pattern on kubernetes where auth tokens are mounted into a volume on a pod.

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - image: nginx
    name: nginx
    volumeMounts:
    - mountPath: /var/run/secrets/tokens
      name: vault-token
  serviceAccountName: build-robot
  volumes:
  - name: vault-token
    projected:
      sources:
      - serviceAccountToken:
          path: vault-token
          expirationSeconds: 7200
          audience: vault
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,15 @@ RUN go build -o /grpcurl \
     -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" \
     ./cmd/grpcurl
 
+FROM alpine:3 as alpine
+WORKDIR /
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=builder /etc/passwd /etc/passwd
+COPY --from=builder /grpcurl /bin/grpcurl
+USER grpcurl
+
+ENTRYPOINT ["/bin/grpcurl"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -16,6 +16,15 @@ RUN go build -o /grpcurl \
     -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" \
     ./cmd/grpcurl
 
+FROM alpine:3 as alpine
```
dragonsinth, line 19: what is alpine:3 in particular?

dragonsinth:
seems reasonable to me

Dockerfile
BenRedmondBenham: @dragonsinth  This will use the latest stable version of alpine without jumping up major versions.

dragonsinth (approved):
Looks like copy/paste from the `scratch` version so LGTM

PR #316, opened 2022-06-12
Feature: Bazel Support

sgammon:
Hey there Fullstory,

I have no idea if this is useful for you guys, but I added Bazel support in a fork because we need it in our environment. It's a pretty conventional add with Gazelle, and should be maintainable along those lines as well. If there is any interest, I would be happy to contribute it back :)

#### Changes enclosed

- [x] Add support for Gazelle and run it
- [x] Add baseline dependencies like proto and `rules_python`
- [x] Add support for external Bazel repos
- [x] Generate `deps.bzl`
- [x] Apply patches for transitive modules that don't build
- [x] Update to Go 1.18
- [ ] Known issues
  - [ ] The tests don't yet run

### To build/run with Bazel

There are two ways to build and run, the way things are set up with Gazelle. There is, the (1) build/run from source way (which you should use during development), and the (2) run from bindist way (which downstream Bazel codebases should use).

Provided for easy copy/paste:
```
git clone git@github.com:sgammon/grpcurl.git -b bazel && cd ./grpcurl
```

**Method 1: Local build/run**
```bash
bazel build //:grpcurl
```

```bash
bazel run //:grpcurl -- -help
```

**Method 2: Downstream run**

First, the user adds the bindist to their own `WORKSPACE`, with:
```starlark
workspace(
  # ...
)

grpcurl_version = "b8e91f6f5b182300c1bf1d3c3f5bf2c221d826cf"  # (or a release, like `v1.8.6`)

http_archive(
    name = "com_github_fullstorydev_grpcurl",
    sha256 = "40819664a05748f6af6cdc814ea70ca09b2a4e3b9d5f5515d459c43cb1e3c55d",
    strip_prefix = "grpcurl-%s" % grpcurl_version,
    url = "https://github.com/sgammon/grpcurl/archive/%s/archive.tar.gz" % grpcurl_version,
)

load("@com_github_fullstorydev_grpcurl//:grpcurl_repositories.bzl", "grpcurl_bindist_repository")

grpcurl_bindist_repository(
    name = "grpcurl",
    version = "1.8.6",
)
```

Then, they can run it in their own Bazel project easily with:
```
bazel run @grpcurl//:grpcurl
```


sgammon:
Update: I've pushed the same support for `grpcui` at fullstorydev/grpcui#188

sgammon:
I should also say, I'm happy to contribute a CI job, or Renovate config, etc., to reduce any potential maintenance burden in merging this. Thanks again for such an awesome tool!

PR #323, opened 2022-07-15
update the go version to 1.18

kapilsharma7:
in this branch I have update the go version to 1.18 as the previous version is only supported the go version 1.15 to 1.17


kapilsharma7:
this is required as some one is using update go version than they will see an error related to go version

dragonsinth:
Why are you trying to vendor in 1300 files???

kapilsharma7:
As I changed the dependency in the mod file, the compiler showed some code error and gave the error message to run the go mod vendor again and again until I run the command go mod vendor, I checked the package unit test cases after running the command and everything is working fine with the go version 1.18.

dragonsinth:
Umm... yeah there's no way we're going to check vendored 1300 files into this repo.

I don't even see a compelling reason to update this to Go 1.18.  Won't that just force downstream consumers to use Go 1.18?   Using the oldest Go version that works makes it easier for consumers.

Issue #300, opened 2022-03-28
Update to Go 1.18


scotthew1:
It looks like #250 has been open for awhile now, but 1.18 has a particularly enticing change for macOS users.

From the [change log](https://tip.golang.org/doc/go1.18):

> [crypto/x509](https://tip.golang.org/pkg/crypto/x509)
> 
> [Certificate.Verify](https://tip.golang.org/pkg/crypto/x509/#Certificate.Verify) now uses platform APIs to verify certificate validity on macOS and iOS when it is called with a nil [VerifyOpts.Roots](https://tip.golang.org/pkg/crypto/x509/#VerifyOpts.Roots) or when using the root pool returned from [SystemCertPool](https://tip.golang.org/pkg/crypto/x509/#SystemCertPool).

nehalshah50:
Same here. We are getting go package vulnerability CVE-2021-36331 as well. Please upgrade ASAP

scaswell-tsys:
Is there a plan to have this upgrade done soon? I've had to remove grpcurl from my docker image due to a dozen or so vulnerabilities in go in versions prior to 1.18.1.

dragonsinth:
Just to be clear on the ask, you want the released binaries and images to be built with Go 1.18 so that the binaries have the CVEs addressed?  You're not asking for a `go.mod` minimum version bump.

scaswell-tsys:
Yes, I'm asking for the released binaries and images to be built on Go 1.18
to address the CVE.

On Wed, Jul 20, 2022 at 2:49 PM Scott Blum ***@***.***> wrote:

> Just to be clear on the ask, you want the released binaries and images to
> be built with Go 1.18 so that the binaries have the CVEs addressed? You're
> not asking for a go.mod minimum version bump.
>
> —
> Reply to this email directly, view it on GitHub
> <https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_fullstorydev_grpcurl_issues_300-23issuecomment-2D1190634522&d=DwMCaQ&c=Z4P52L0foFKAY1wcP-GmiQ&r=CfJc7E9Y5rcraXO6S6e5t3G1fc3xkM5N1A8qyKvsdys&m=f3yKN49C_yA9y0v120FZKvm0u0NWDTjM0JWE05jei4R_4ERMnf9IMpUD-bMQ6Upz&s=y2C0Dk-8XZzvmfXSUG1UyHJEK6cOq0KGluVlTw_ja7k&e=>,
[...]

PR #325, opened 2022-07-20
Add go 1.18 support; set Dockerfile to go 1.18

dragonsinth:
https://github.com/fullstorydev/grpcurl/issues/300

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,12 @@ shared_configs:
         name: Run tests
         command: |
           make deps test
-
+  full_job_steps: &full_job_steps
+    - checkout
+    - run:
+        name: Run tests
+        command: |
+          make ci
 
 # Use the latest 2.1 version of CircleCI pipeline process engine. See: https://circleci.com/docs/2.0/configuration-reference
 version: 2.1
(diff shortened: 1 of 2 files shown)


dragonsinth:
@gpassini @sanjaybv @franklinlindemberg

.circleci/config.yml
```diff
@@ -5,46 +5,38 @@ shared_configs:
...
     docker:
-      - image: circleci/golang:1.16
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - go-mod-v4-{{ checksum "go.sum" }}
-      - run:
-          name: Install Dependencies
-          command: go mod download
```
gpassini, line 30: We don't need to run this command anymore?

dragonsinth: Nah, the necessary modules will be downloaded when it tries to build.  This was broken out before in order to try to populate Go's module cache and then persist it to Circle's cache.  I found this to be a net loss; restoring the cache could take several minutes, while the non-cached versions would completely in like 1 minute.

gpassini (approved):
LGTM!

.circleci/config.yml
```diff
@@ -5,46 +5,38 @@ shared_configs:
...
-      - run:
-          name: Run tests
-          command: |
-            #mkdir -p /tmp/test-reports
-            #gotestsum --junitfile /tmp/test-reports/unit-tests.xml
-            make ci
-    #- store_test_results:
-    #    path: /tmp/test-reports
+      - image: cimg/go:1.16
+    steps: *full_job_steps
```
franklinlindemberg, line 28: do you know why only build-1-16 is using full_job_steps?

franklinlindemberg (approved):
LGTM

Will you do a release with this new version?

.circleci/config.yml
dragonsinth: I don't; I get the idea of only running the full deal on 1/4 steps, but I'm not sure why Go 1.16 was specifically chosen for this.

dragonsinth:
> Will you do a release with this new version?

Yeah, once I figure out how to do that :D

franklinlindemberg:
@dragonsinth You've probably seem this, but https://github.com/fullstorydev/grpcurl/tree/master/releasing describes the release doc.

I think it expects docker to be running, so we'll need some tweak. Also do you have access to our docker hub? It will be needed to push the image

dragonsinth:
> @dragonsinth You've probably seem this, but https://github.com/fullstorydev/grpcurl/tree/master/releasing describes the release doc.

Yeah, it's a bit of a read tho.  I need to be sure it's up to date.

> I think it expects docker to be running, so we'll need some tweak. Also do you have access to our docker hub? It will be needed to push the image

My plan is to run through it in a Linux VM.  I probably don't have access to docker hub, do you know anything about that?  I'd love to pair up on this-- has anyone other than @jhump ever run through the process before?

Issue #333, opened 2022-09-01
Add support for ALTS authentication.


jayeye:
Is anyone working on this? If not, I'll do it.

derpferd:
Also needed this feature so just added it and opened a PR. 😄

PR #341, opened 2022-10-22
Added alts credential option

derpferd:
Added flag that sets up ALTS credentials to connect to gRPC services using Application Layer Transport Security.

Reference: https://grpc.io/docs/languages/go/alts/
Fixes: #333


cmd/grpcurl/grpcurl.go
```diff
@@ -407,7 +422,14 @@ func main() {
 			opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(*maxMsgSz)))
 		}
 		var creds credentials.TransportCredentials
-		if !*plaintext {
+		if *plaintext {
+			if *authority != "" {
+				opts = append(opts, grpc.WithAuthority(*authority))
+			}
+		} else if *usealts {
+			creds = alts.NewClientCreds(alts.DefaultClientOptions())
```
jhump, line 430: Would we not want to support some options here, like specifying the trusted/expected server service account or allowing a custom handshaker address?

derpferd: Good point, I don't use these arguments but will add in case someone needs to.

derpferd: Added in new commit

cmd/grpcurl/grpcurl.go
```diff
@@ -203,6 +205,12 @@ func init() {
 		-use-reflection is used in combination with a -proto or -protoset flag,
 		the provided descriptor sources will be used in addition to server
 		reflection to resolve messages and extensions.`))
+	flags.Var(&altsTargetServiceAccounts, "alts-target-service-account", prettify(`
+	    A list of expected target service accounts. If the service is running as
+		any of the provided service account the connection and request will be
+		made else the connection will fail to connect. If not target service
+		accounts are provided, no client side authorization will be done and any
+		connection will be allowed.`))
```
jhump, line 213:
```suggestion
		The full email address of the service account that the server is
		expected to be using when ALTS is used. You can specify this option
		multiple times to indicate multiple allowed service accounts. If the
		server authenticates with a service account that is not one of the
		expected accounts, the RPC will not be issued. If no such arguments are
		provided, no check will be performed, and the RPC will be issued
		regardless of the server's service account.
```

jhump (commented):
Full disclosure: I am no longer a committer for this repo. So you'll need someone like @jameremo or @dragonsinth to review and approve this change before it can be merged.

But since it doesn't add any dependencies, it seems reasonable. (If accepted, a similar change should likely be made to [grpcui](https://github.com/fullstorydev/grpcui), too.)

Itsindigo:
@jameremo @dragonsinth any possibility of merging this?

cmd/grpcurl/grpcurl.go
```diff
@@ -290,6 +306,21 @@ func main() {
...
+	}
+	if *usealts && *key != "" {
+		fail(nil, "The -alts and -key arguments are mutually exclusive.")
+	}
+	if *altsHandshakerServiceAddress != "" && !*usealts {
+		fail(nil, "The -alts-handshaker-service argument must be used with the -alts argument.")
+	}
+	if len(altsTargetServiceAccounts) > 0 && !*usealts {
+		fail(nil, "The -alts-target-service-account argument must be used with the -alts argument.")
+	}
```
dragonsinth, line 323: This is kind of a combinatorical mess.  Perhaps early up we could assign `usetls := !*plaintext && !*usealts`?

Then you could collapse cases like `*plaintext && *cert` and `*usealts && *cert` into something like `if *cert && !usetls { ... }`

cmd/grpcurl/grpcurl.go
```diff
@@ -407,7 +438,21 @@ func main() {
...
+			clientOptions := alts.DefaultClientOptions()
+			if len(altsTargetServiceAccounts) > 0 {
+				clientOptions.TargetServiceAccounts = altsTargetServiceAccounts
+			}
+			if *altsHandshakerServiceAddress != "" {
+				clientOptions.HandshakerServiceAddress = *altsHandshakerServiceAddress
+			}
+			creds = alts.NewClientCreds(clientOptions)
+		} else {
+			// Use TLS
```
dragonsinth, line 455: Maybe...
```suggestion
		} else if usetls {
```

Then panic if you fall through to a final `else`

cmd/grpcurl/grpcurl.go
```diff
@@ -55,7 +56,11 @@ var (
 		Print version.`))
 	plaintext = flags.Bool("plaintext", false, prettify(`
 		Use plain-text HTTP/2 when connecting to server (no TLS).`))
-	insecure = flags.Bool("insecure", false, prettify(`
+	usealts = flags.Bool("alts", false, prettify(`
+		Use Application Layer Transport Security (ALTS) when connecting to server.`))
+	altsHandshakerServiceAddress = flags.String("alts-handshaker-service", "", prettify(`If set, this server will be used to do the ATLS handshaking.`))
+	altsTargetServiceAccounts    multiString
```
dragonsinth, line 62: Maybe move the new stuff down to line 74.5?

I also wouldn't be opposed to adding a few line breaks to visually separate clusters of related options

cmd/grpcurl/grpcurl.go
derpferd: Done

cmd/grpcurl/grpcurl.go
derpferd: Done

derpferd:
@dragonsinth ready for your review again. Applied the requested changes.

dragonsinth (approved):
Thanks, looks great

PR #360, opened 2023-01-13
feat: add -keep-proto-names flag

clement-sciascia:
This flag changes the JSON output, keeping the original fields' name as declared in proto, instead of JSON's default lowerCamelCase format

Signed-off-by: Clement Sciascia <clement.sciascia@ovhcloud.com>


anhnmt:
I think this feature is necessary

anhnmt (approved):
LGTM

Issue #193, opened 2020-10-20
How to send and receive "google.protobuf.Any"?


maxliu007:
Here is the protobuf message:
message MergeResourcesRequest {
  ResourceType type = 1;
  repeated google.protobuf.Any resources = 2;
}

How can we fire this message to ther server with grpcurl? As it reports follow error:
Error invoking method "utsv.event.api.v1.EventService/MergeResources": error getting request data: unknown message type "utsv.event.api.v1.resources.Merchant"

Here ist eh json:
{
	"type" : 15,
	"resources" : [
		{
			"@type": "type.googleapis.com/utsv.event.api.v1.resources.Merchant",
			"id" : 123,
			"code" : "Merchant",
			"name": {
				"en": "desc en",
				"zh": "desc tw"
			}
		}
	]
}

jhump:
@maxliu007, are you using server reflection? If so, then if the server knows about that message type, it should work.

If you are not using server reflection, then you must provide `-proto` or `-protoset` arguments that indicate not only the source for the service you are using but also the sources for all message types.

If you _are_ using server reflection but it's still not working, then there may be an issue with the server reflection implementation in your server. But a work-around could still be to point `grpcurl` at local proto source files for the messages. You can do this, while still using server reflection, but providing _both_ `-proto` arguments (to point to proto sources for messages) _and_ the `-use-reflection` argument. With this, `grpcurl` will try server reflection first, to resolve any messages or extensions, and then, failing that, try the proto sources.

maxliu007:
We are using Java 13 and with server reflection enabled. 
Tested with  following grpc version:
<grpc.version>1.33.0</grpc.version>
<protoc.version>3.12.0</protoc.version>

When we are testing the non-Any message, grpculr can reply positive result. However it fails whenever we are firing the Any message to the server. From what we observe so far, it has been stucked on the grpcurl.  **So I wonder if grpcurl can support Any message or not.**

Here is the one tested with verbose output:
*****************************************************
c:\temp\sample\merge-resources-merchant.json"

Resolved method descriptor:
rpc MergeResources ( .utsv.event.api.v1.MergeResourcesRequest ) returns ( .utsv.event.api.v1.MergeResourcesResponse );

Request metadata to send:
(empty)
Error invoking method "utsv.event.api.v1.EventService/MergeResources": error getting request data: unknown message type "utsv.event.api.v1.resources.Merchant"
*****************************************************

jhump:
[...]

PR #379, opened 2023-03-20
Show additional error details when an rpc fails

fdfzcq:
Fixes the following issue:
> If an error is returned with additional details together with code and message, grpcurl will format it poorly and not show the additional details.

### DIFF
--- a/format.go
+++ b/format.go
@@ -136,10 +136,25 @@ type Formatter func(proto.Message) (string, error)
 func NewJSONFormatter(emitDefaults bool, resolver jsonpb.AnyResolver) Formatter {
 	marshaler := jsonpb.Marshaler{
 		EmitDefaults: emitDefaults,
-		Indent:       "  ",
 		AnyResolver:  resolver,
 	}
-	return marshaler.MarshalToString
+	// Workaround for indentation issue in jsonpb with Any messages.
+	// Bug was originally fixed in https://github.com/golang/protobuf/pull/834
+	// but later re-introduced before the module was deprecated and frozen.
+	// If jsonpb is ever replaced with google.golang.org/protobuf/encoding/protojson
+	// this workaround will no longer be needed.
+	formatter := func(message proto.Message) (string, error) {
+		output, err := marshaler.MarshalToString(message)
+		if err != nil {
+			return "", err
+		}
+		var buf bytes.Buffer
+		if err := json.Indent(&buf, []byte(output), "", "  "); err != nil {
+			return "", err
+		}
+		return buf.String(), nil
+	}
+	return formatter
 }
 
 // NewTextFormatter returns a formatter that returns strings in the protobuf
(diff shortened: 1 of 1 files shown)


format.go
```diff
@@ -274,11 +289,11 @@ func (r *anyResolver) Resolve(typeUrl string) (proto.Message, error) {
...
-	// populate any extensions for this message, too
-	if exts, err := r.source.AllExtensionsForType(mname); err != nil {
-		return nil, err
-	} else if err := r.er.AddExtension(exts...); err != nil {
-		return nil, err
+	// populate any extensions for this message, too (if there are any)
+	if exts, err := r.source.AllExtensionsForType(mname); err == nil {
+		if err := r.er.AddExtension(exts...); err != nil {
+			return nil, err
+		}
```
dragonsinth, line 296: What's going on here?

fdfzcq: I believe this fixes https://github.com/fullstorydev/grpcurl/issues/193 but @erikjoh can confirm

erikjoh: To give some extra context, we are trying to adopt the richer error model of grpc but grpcurl currently outputs the errors like this:

```
ERROR:
  Code: InvalidArgument
  Message: Request contains invalid fields
  Details:
  1)	{"@error":"google.rpc.BadRequest is not recognized; see @value for raw binary message data","@type":"type.googleapis.com/google.rpc.BadRequest","@value":"CiQKCnN0cmluZ19vbmUSFk5vIGVtcHR5IHZhbHVlIGFsbG93ZWQKJAoKc3RyaW5nX3R3bxIWTm8gZW1wdHkgdmFsdWUgYWxsb3dlZA=="}
```

The problem with the current code here is that when using service reflection to resolve `google.rpc.BadRequest`, it doesn't work since `google.rpc.BadRequest` doesn't have any protobuf extension.
You can see here that `r.source.AllExtensionsForType(mname)` returns an `err` `github.com/jhump/protoreflect/grpcreflect.elementNotFoundError` which then aborts the formatting.

image

With the change in this PR, the `elementNotFoundError` will be ignored and the formatting of the error will successfully be processed and printed:

```
ERROR:
  Code: InvalidArgument
  Message: Request contains invalid fields
  Details:
  1)	{
    	  "@type": "type.googleapis.com/google.rpc.BadRequest",
    	  "field_violations": [
    	    {
    	      "field": "string_one",
    	      "description": "No empty value allowed"
    	    },
    	    {
    	      "field": "string_two",
    	      "description": "No empty value allowed"
    	    }
    	  ]
    	}
```

This new code should add extensions (i.e. `r.er.AddExtension(exts...)`) only if extensions could be found with `r.source.AllExtensionsForType(mname)`.

format.go
```diff
@@ -136,10 +136,25 @@ type Formatter func(proto.Message) (string, error)
...
+		if err != nil {
+			return "", err
+		}
+		var buf bytes.Buffer
+		if err := json.Indent(&buf, []byte(output), "", "  "); err != nil {
+			return "", err
+		}
+		return buf.String(), nil
+	}
+	return formatter
```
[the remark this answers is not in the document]
jhump: Yep, this PR LGTM.

format.go
dragonsinth: Gotcha, that makes sense.  Why not also ignore the error from `r.er.AddExtension(exts...)`?

fdfzcq:
Hello @dragonsinth @jhump can we have this PR merged? :) thx!

dragonsinth (approved):

fdfzcq:
@dragonsinth thank you for merging the PR, could you make a new release too? :)

PR #385, opened 2023-04-17
Add "checkgenerate" make target to CI

jhump:
Basically the same as https://github.com/fullstorydev/grpcui/pull/239.


Makefile
```diff
@@ -31,6 +33,19 @@ docker:
 	docker build -t fullstorydev/grpcurl:$(dev_build_version) .
 	@rm VERSION
 
+.PHONY: generate
+generate: .tmp/protoc/bin/protoc
```
jhump, line 37: Doh. This repo never even had a `make generate` target!

Makefile
```diff
@@ -44,7 +59,7 @@ vet:
 
 .PHONY: staticcheck
 staticcheck:
-	@go install honnef.co/go/tools/cmd/staticcheck@v0.3.3
+	@go install honnef.co/go/tools/cmd/staticcheck@v0.4.3
```
jhump, line 62: I have Go 1.20 installed, and the older version of staticcheck was blowing up when it happened upon 1.20 object files. So I went ahead and updated it.

download_protoc.sh
```diff
@@ -0,0 +1,32 @@
...
+  PROTOC_ARCH="x86_64"
+fi
+
+PROTOC="${PWD}/.tmp/protoc/bin/protoc"
+
+if [[ "$(${PROTOC} --version 2>/dev/null)" != "libprotoc 3.${PROTOC_VERSION}" ]]; then
+  rm -rf ./.tmp/protoc
+  mkdir -p .tmp/protoc
+  curl -L "https://github.com/google/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-${PROTOC_OS}-${PROTOC_ARCH}.zip" > .tmp/protoc/protoc.zip
+  cd ./.tmp/protoc && unzip protoc.zip && cd ..
```
dragonsinth, line 30:
```suggestion
  cd ./.tmp/protoc && unzip protoc.zip && cd ../..
```

dragonsinth (approved):
LGTM with path nit

internal/testing/cmd/bankdemo/bank.go
```diff
@@ -12,6 +12,7 @@ import (
 
 // bankServer implements the Bank gRPC service.
 type bankServer struct {
+	UnimplementedBankServer
```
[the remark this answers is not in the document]
jhump: Yeah, it stinks. But that's how the latest version of grpc-go works -- this is required and there's a compile error without it. 🤷

download_protoc.sh
jhump: Incorporated same changes in this PR as from the [grpcui one](https://github.com/fullstorydev/grpcui/pull/239). So this is using `pushd`/`popd` now.

dragonsinth (approved):

PR #395, opened 2023-05-12
Fix nil-dereference panic that can happen invoking a server stream

### LINKED ISSUE #394: nil pointer dereference panic when accessing a streaming server that is not implemented
I believe the `invokeServerStream` function does not properly check the `err` on [line 265](https://github.com/fullstorydev/grpcurl/blob/master/invoke.go#L265).

This (can) cause a nil panic if the method is not implemented (or another error occurs) on [line 267](https://github.com/fullstorydev/grpcurl/blob/master/invoke.go#L267)'s `str.Header()`.

---

**Update:** Interestingly I'm only seeing this behavior when using `github.com/fullstorydev/grpchan`. I'm still investigating but I think before a PR is opened the question needs to be asked as to which the proper behavior is. That is to say: How should a Streaming service note that it's not implemented?

This isn't a huge deal as I immediately plan on implementing this service but it was a fun rabbit hole to go down so far, so I'll keep investigating before opening a PR.

---

~~I believe this is as simple as following the pattern used in the `invokeUnary` function above and am preparing a PR.~~

Here is an example (trimmed) trace:

```
invalid memory address or nil pointer dereference goroutine 5 [running]:
panic({0x171fb20, 0x1e92260})
	/usr/local/Cellar/go/1.20.2/libexec/src/runtime/panic.go:884 +0x213
github.com/jhump/protoreflect/dynamic/grpcdynamic.(*ServerStream).Header(...)
	/Users/fernferret/projects/wab/vendor/github.com/jhump/protoreflect/dynamic/grpcdynamic/stub.go:178
github.com/fullstorydev/grpcurl.invokeServerStream({0x1912408, ...
	/Users/fernferret/projects/wab/vendor/github.com/fullstorydev/grpcurl/invoke.go:267 +0x217
github.com/fullstorydev/grpcurl.InvokeRPC({0x1912408, ...
	/Users/fernferret/projects/wab/vendor/github.com/fullstorydev/grpcurl/invoke.go:147 +0x78a
github.com/fullstorydev/grpcui.invokeRPC({0x1912408, ...
	/Users/fernferret/projects/wab/vendor/github.com/fullstorydev/grpcui/handlers.go:464 +0x69c
github.com/fullstorydev/grpcui.RPCInvokeHandlerWithOptions.func1({0x1910cf0, ...
	/Users/fernferret/projects/wab/vendor/github.com/fullstorydev/grpcui/handlers.go:98 +0x3c7
net/http.HandlerFunc.ServeHTTP(0x171abe0?, ...
	/usr/local/Cellar/go/1.20.2/libexec/src/net/http/server.go:2122 +0x2f
```

Thanks so much for this great library!

jhump:
This should resolve #394.

The bidi and client streaming methods already had stream accesses behind a `str != nil` check. It was just server-side streaming that was missing these checks :(

### DIFF
--- a/invoke.go
+++ b/invoke.go
@@ -264,8 +264,10 @@ func invokeServerStream(ctx context.Context, stub grpcdynamic.Stub, md *desc.Met
 	// Now we can actually invoke the RPC!
 	str, err := stub.InvokeRpcServerStream(ctx, md, req)
 
-	if respHeaders, err := str.Header(); err == nil {
-		handler.OnReceiveHeaders(respHeaders)
+	if str != nil {
+		if respHeaders, err := str.Header(); err == nil {
+			handler.OnReceiveHeaders(respHeaders)
+		}
 	}
 
 	// Download each response message
@@ -288,7 +290,9 @@ func invokeServerStream(ctx context.Context, stub grpcdynamic.Stub, md *desc.Met
 		return fmt.Errorf("grpc call for %q failed: %v", md.GetFullyQualifiedName(), err)
 	}
 
-	handler.OnReceiveTrailers(stat, str.Trailer())
+	if str != nil {
+		handler.OnReceiveTrailers(stat, str.Trailer())
+	}
 
 	return nil
 }


jhump:
@dragonsinth or @sanjaybv 👀

dragonsinth (approved):

PR #407, opened 2023-08-25
Use `grpc.reflection.v1.ServerReflection`

sigv:
The proper `v1` gRPC reflection has been around for a bit. The "client auto" mechanism still supports falling back to `v1alpha`. Not many support the `v1`, but we should default to it when possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ This program accepts messages using JSON encoding, which is much more friendly f
 humans and scripts.
 
 With this tool you can also browse the schema for gRPC services, either by querying
-a server that supports [server reflection](https://github.com/grpc/grpc/blob/master/src/proto/grpc/reflection/v1alpha/reflection.proto),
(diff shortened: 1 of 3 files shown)


sigv:
Just to make sure it's clear: How it works in practice is, it tries the `v1` endpoint, which is newer. If `v1` fails, then it tries `v1alpha` essentially resulting in already existing behavior/backwards compatibility.

jhump:
👍 

This was something I added to `grpcreflect` a while ago (nearly a [year ago](https://github.com/jhump/protoreflect/pull/535)), after seeing that the v1 version of the proto had been around quite a while and that Connect was [already using it](https://github.com/connectrpc/grpcreflect-go/commit/012dfb012bcc20406dfcb6f43f10d645de6ba57e).

Sadly, at that time, none of the official implementations actually provided v1, only v1alpha (so I filed [an issue](https://github.com/grpc/grpc-go/issues/5684)). It was finally implemented in the grpc-go module [recently](https://github.com/grpc/grpc-go/pull/6329/files) (your welcome 🤣).

dragonsinth:
LOL, you go @jhump !

Issue #414, opened 2023-09-26
SIGSEGV: panic: runtime error: invalid memory address or nil pointer dereference in v1.8.8


mprimeaux:
I'm using an existing `.proto` file that been working nicely with `grpcurl` for the past months. It appears that version `1.8.8` released a few days ago now results in a panic.

I downloaded version `1.8.7` and things work as expected. 

```shell
❯ cat mutate/challenge.action.add.json | ~/Downloads/grpcurl -plaintext -vv -d @ -format=json -use-reflection=false -proto ../internal/api/hydra_service.proto localhost:50503 api.Hydra.Mutate

Resolved method descriptor:
// Mutates one or more underlying data structures.
rpc Mutate ( .api.MutateRequest ) returns ( .api.MutateResponse );

Request metadata to send:

Response headers received:
content-type: application/grpc

Estimated response size: 50 bytes

Response contents:
{
  "data": [
    {
          "id": "6789a357-0d07-4cf9-8025-2438d7e639f2"
        }
  ]
}

Response trailers received:
(empty)
Sent 1 request and received 1 response
```

Using version `1.8.8` results in the following panic:

```shell
❯ cat mutate/challenge.action.add.json |  -plaintext -vv -d @ -format=json -use-reflection=false -proto ../internal/api/hydra_service.proto localhost:50503 api.Hydra.Mutate
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x28 pc=0x100f714e8]

goroutine 1 [running]:
github.com/jhump/protoreflect/desc/protoparse.parseToProtoRecursive({0x101406c80, 0x14000195308}, {0x140001bf8e0, 0x1c}, 0x1400025f608?, 0x100f08bf4?, 0x0?)
	github.com/jhump/protoreflect@v1.15.2/desc/protoparse/parser.go:389 +0x148
github.com/jhump/protoreflect/desc/protoparse.parseToProtoRecursive.func1(0x14000365380, 0x140003b2eb0, 0x140001e3810, {0x101406c80, 0x14000195308}, 0x1013ba560?, 0x140001d9f01?)
	github.com/jhump/protoreflect@v1.15.2/desc/protoparse/parser.go:401 +0x164
[...]

PR #416, opened 2023-10-03
SIGSEGV: panic: runtime error: invalid memory address or nil pointer dereference in protoreflect

mprimeaux:
This PR addresses issue #414. 

- Executed the updatedeps make target
- protoreflect upgraded to v1.15.3, which addresses this panic.


dragonsinth:
@mprimeaux I think a more targeted PR probably better; this one doesn't build on Go 1.18 due to some of the updates.

mprimeaux:
@dragonsinth I updated the `protoreflect` dependency from `v1.15.2` to `v1.15.3`.

dragonsinth (approved):

Issue #404, opened 2023-07-31
New release? 


nordicdyno:
Can you create a new release, please? The latest one doesn't work with new google proto options like `debug_redact`, however grpcurl installed from main branch have no such issue.

gfrankliu:
The security scanner also reports lots of vulnerabilities in the released binaries since they were compiled using the old vulnerable go compiler.

dragonsinth:
It's in the works!

gfrankliu:
Thanks for the release!

I noticed one difference in the binary release. 1.8.7 was probably built was cgo disabled, but 1.8.8 has C lib dependencies:

I tried linux_x86_64.tar.gz, for 1.8.7
```
ldd grpcurl
	not a dynamic executable
```
and it works fine in the alpine docker image

for 1.8.8
```
Error loading shared library libresolv.so.2: No such file or directory (needed by ./grpcurl.new)
Error relocating ./grpcurl.new: __res_search: symbol not found
```

It works on Debian:
```
$ ldd grpcurl
	linux-vdso.so.1 (0x00007ffd2250a000)
	libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f7bd5383000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f7bd537e000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f7bd519c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7bd53ad000)
```

Can the binary releases be compiled with cgo off, so they can have best compatibility?

gfrankliu:
I checked the "grcpurl" binary inside the container image https://hub.docker.com/r/fullstorydev/grpcurl/tags and see it doesn't have C dependencies. 
This is confirmed by [this line](https://github.com/fullstorydev/grpcurl/blob/master/Dockerfile#L13) in the Dockerfile.
Maybe we should add "CGO_ENABLED=0" in the Makefile so that future releases won't have this problem, eg:
```
dev_build_version=$(shell git describe --tags --always --dirty)

export PATH := $(shell pwd)/.tmp/protoc/bin:$(PATH)

export PROTOC_VERSION := 22.0

export CGO_ENABLED := 0

...
```

dragonsinth:
[...]

PR #420, opened 2023-10-24
Disable CGO for improved compatibility across distros

gpassini:
This PR disables CGO so we recover the same behaviour we had in v1.8.7, where we don't have any dependencies on C libraries expected to be found in the environment, and was raised in https://github.com/fullstorydev/grpcurl/issues/404#issuecomment-1762593336.

Tested by running `make install` before and after the changes and inspecting the produced binary from a Linux machine (Debian in this case).
**Before**
```sh
ldd $(which grpcurl)
	linux-vdso.so.1 (0x00007ffe121e7000)
	libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f2346e02000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2346de0000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2346c0c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2346e26000)
```
**After**
```sh
ldd $(which grpcurl)
	not a dynamic executable
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,8 @@ errcheck:
 
 .PHONY: test
 test:
-	go test -race ./...
+	# The race detector requires CGO: https://github.com/golang/go/issues/6508
+	CGO_ENABLED=1 go test -race ./...
 
 .tmp/protoc/bin/protoc: ./Makefile ./download_protoc.sh
 	./download_protoc.sh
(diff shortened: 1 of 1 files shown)


dragonsinth (approved):
Maybe a similar change in grpcui as well?

dragonsinth:
#420 🔥

PR #424, opened 2023-10-26
Enable xDS credentials

PapaCharlie:
This change should be relatively straightforward. It is a noop outside of the context of xDS (as demonstrated by the fact that the tests all pass), but it enables xDS-provided certificates (i.e. the ones that would be provided/specified in GRPC_XDS_BOOTSTRAP). See proposal [A29](https://github.com/grpc/proposal/blob/master/A29-xds-tls-security.md#go) for additional detail.


grpcurl.go
```diff
@@ -629,7 +630,16 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 			TransportCredentials: creds,
 			writeResult:          writeResult,
 		}
+	} else {
+		creds = insecure.NewCredentials()
+	}
+
+	var err error
+	creds, err = xdsCredentials.NewClientCredentials(xdsCredentials.ClientOptions{FallbackCreds: creds})
```
jhump, line 638: It seems like this code path should only be triggered if the `address` includes the `xds://` scheme, no?

PapaCharlie: Actually, this is what I mentioned in my PR description. This is a noop if the target _isn't_ `xds:///`, so we can avoid the potentially awkward check for the prefix. I can add it if you want to have it explicitly

PapaCharlie: I updated this to only do it if it's an `xds:///` target

dragonsinth:
This probably needs a rebase.  Are you sure that the XDS creds should wrap the errSignalingCreds and not the other way around?  The code flow is a little tricky to absorb.

PapaCharlie:
Let me check

PapaCharlie:
I think you're right, it's likely best to wrap the final `creds` with the `errSignalingCreds`.

PapaCharlie:
Yeah I confirmed that this works! Since `errSignalingCreds` embeds the `credentials.TransportCredentials`, it still implements the [`UsesXDS`](https://github.com/grpc/grpc-go/blob/v1.57.1/xds/internal/resolver/xds_resolver.go#L112) interface that's required by the code.

dragonsinth:
> Yeah I confirmed that this works! Since `errSignalingCreds` embeds the `credentials.TransportCredentials`, it still implements the [`UsesXDS`](https://github.com/grpc/grpc-go/blob/v1.57.1/xds/internal/resolver/xds_resolver.go#L112) interface that's required by the code.

Waaaait a second.  That's definitely not true.

dragonsinth:
I had to open another PR to get CI to work, for some reason it's getting stuck on this one.
https://github.com/fullstorydev/grpcurl/pull/441

WDYT?

PapaCharlie:
> > Yeah I confirmed that this works! Since `errSignalingCreds` embeds the `credentials.TransportCredentials`, it still implements the [`UsesXDS`](https://github.com/grpc/grpc-go/blob/v1.57.1/xds/internal/resolver/xds_resolver.go#L112) interface that's required by the code.
> 
> Waaaait a second. That's definitely not true.

Why not? This prints `true`: https://go.dev/play/p/TxQSrKPngtc?v=

dragonsinth:
> Why not? This prints `true`: https://go.dev/play/p/TxQSrKPngtc?v=

That one works because you embedded  the concrete fooBar, which implements both interfaces.  If you change it to this, it doesn't work:

https://go.dev/play/p/OyIwdHFyJ3x

(I had to rename the interfaces so that the interface names and method names dont clash)

dragonsinth (approved):

PR #428, opened 2023-10-31
Added initial support for -t flag to show timings

KaibutsuX:
Shows very basic timing data for the Dial stage (TLS setup and BlockingDial) and InvokeRPC method as well as the total time.

### DIFF
--- a/cmd/grpcurl/grpcurl.go
+++ b/cmd/grpcurl/grpcurl.go
@@ -292,6 +292,9 @@ func (d *timingData) Child(title string) *timingData {
 }
 
 func (d *timingData) Done() {
+	if d == nil {
+		return
+	}
 	if d.Value == 0 {
 		d.Value = time.Since(d.Start)
 	}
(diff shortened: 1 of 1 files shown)


dragonsinth (commented):
Could we reduce the options by just making this part of `-vv`?  We could just log e.g. "start Dial" -> "end Dial (35ms)" in the places where this PR is capturing time.Now() and time.Since()

dragonsinth:
How much are we talking about? I assumed this would be kinda coarse grained
timing on major steps.

KaibutsuX:
I haven't dug into what further information I can get out of the connection, I assumed there might be finer grained details. But for now this at least gets me the information I was looking for to debug some slow responses from the a server. If you want this in -vv I can move it there.

dragonsinth:
@KaibutsuX I pushed you a commit to clean this up a bit: WDYT?

dragonsinth (approved):

PR #435, opened 2023-12-08
Expand the documentation of -max-time to clarify this sets the RPC timeout

dimo414:
I was looking for how to specify a RPC timeout or [deadline](https://grpc.io/docs/guides/deadlines/) via `grpcurl` and tried grepping the `--help` output, but only matched `-connect-timeout`. I found `-max-time` by browsing the codebase looking for `.WithTimeout()` calls. So the motivation of this PR is both to generally clarify the behavior of this flag and also add the words "timeout" and "deadline" specifically to this help text to aid discovery.


dragonsinth (approved):
thanks!

PR #441, opened 2024-01-24
Papa charlie/master

dragonsinth:


grpcurl.go
```diff
@@ -697,3 +707,8 @@ func (c *errSignalingCreds) ClientHandshake(ctx context.Context, addr string, ra
 	}
 	return conn, auth, err
 }
+
+func (c *errSignalingCreds) UsesXDS() bool {
+	xc, ok := c.TransportCredentials.(interface{ UsesXDS() bool })
+	return ok && xc.UsesXDS()
+}
```
PapaCharlie, line 714: I don't think this is necessary. I checked through it with a debugger, `errSignalingCreds` will implement `UsesXDS` as long as the `TransportCredentials` does

PapaCharlie (changes requested):

PapaCharlie:
Oh also, it seems the `UsesXDS` interface disappeared in HEAD of go-grpc anyway, so this is totally moot!

dragonsinth:
> Oh also, it seems the `UsesXDS` interface disappeared in HEAD of go-grpc anyway, so this is totally moot!

So we're on 1.59.0.  I scanned through the gRPC lib and it _looks_ like all of the existing uses even in that version are server-side only.  So I think maybe we don't need this on the client at all.  Can you try it out without:?

grpcurl.go
PapaCharlie: Nevermind, you are totally right, you do need to explicitly implement the interface. The part I don't like about this in general though is that that is completely hidden in the code. It feels strange that that's how it's wired in. In the later versions of grpc-go, are there new interfaces like this?

dragonsinth: No clue.  See my other comment tho-- it looks to me like grpclib only cares about `UsesXDS` on the server side transport.  We're only using client, so I _think_ we could just rip this out.  But you'd need to test it for me.

PR #461, opened 2024-05-06
Add 32bit arm build targets

philippgl:
I want to use grpcurl on a raspberry pi 0 and therefore I need 32 bit binaries. As this might be interesting for others, I thought, I could just create a PR for it.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -8,14 +8,23 @@ builds:
     goarch:
       - amd64
       - 386
+      - arm
       - arm64
       - s390x
       - ppc64le
+    goarm:
+      - 5
+      - 6
+      - 7
     ignore:
       - goos: darwin
         goarch: 386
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -8,14 +8,23 @@ builds:
...
       - amd64
       - 386
+      - arm
       - arm64
       - s390x
       - ppc64le
+    goarm:
+      - 5
+      - 6
+      - 7
```
dragonsinth, line 18: do we need all 3? just thinking about keeping the number of permutations reasonable

philippgl: I only need armv6, but I think armv7 is also quite popular.
This should add exactly 3 builds: armv5, armv6 and armv7 for linux.

dragonsinth (approved):

PR #475, opened 2024-07-10
feat: Add functionality to export proto files

### LINKED ISSUE #206: Proto generation
Hi, can we export proto from grpc server reflection? 

Thanks

### LINKED ISSUE #474: Add --proto-out-dir option to generate .proto files
**Feature Description:**

Add a new option **--proto-out-dir** to grpcurl that allows users to generate .proto files for listed, described, or invoked elements and their transitive dependencies.

**Functionality:**

When using the list and describe verbs, the listed or described elements and their transitive dependencies will be written as .proto files in the specified directory.

**Proposed Implementation:**
Add a new flag:
```go
protoOut = flags.String("proto-out-dir", "", "The name of a directory where the generated .proto files will be written.")
```

**Use Case:**
This feature will be useful for users who need to generate .proto files from gRPC services, especially when working with services that don't have readily available .proto files or when dealing with dynamically generated services.

Would you like me to modify or expand on any part of this issue?

Eitol:
**Feature Description:**

Add a new option **--proto-out-dir** to grpcurl that allows users to generate .proto files for listed, described, or invoked elements and their transitive dependencies.

**Functionality:**

When using the list and describe verbs, the listed or described elements and their transitive dependencies will be written as .proto files in the specified directory.

**Proposed Implementation:**
Add a new flag:
```go
protoOut = flags.String("proto-out-dir", "", "The name of a directory where the generated .proto files will be written.")
```

**Use Case:**
This feature will be useful for users who need to generate .proto files from gRPC services, especially when working with services that don't have readily available .proto files or when dealing with dynamically generated services.

Tested with the starlink gRPC server.

image

TEST: OK 

image

closes #474 
closes #206

### DIFF
--- a/desc_source.go
+++ b/desc_source.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/jhump/protoreflect/desc/protoprint"
 	"io"
 	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/golang/protobuf/proto" //lint:ignore SA1019 we have to import this because it appears in exported API
@@ -258,19 +260,9 @@ func reflectionSupport(err error) error {
 // given output. The output will include descriptors for all files in which the
 // symbols are defined as well as their transitive dependencies.
 func WriteProtoset(out io.Writer, descSource DescriptorSource, symbols ...string) error {
-	// compute set of file descriptors
-	filenames := make([]string, 0, len(symbols))
-	fds := make(map[string]*desc.FileDescriptor, len(symbols))
-	for _, sym := range symbols {
-		d, err := descSource.FindSymbol(sym)
-		if err != nil {
-			return fmt.Errorf("failed to find descriptor for %q: %v", sym, err)
-		}
-		fd := d.GetFile()
-		if _, ok := fds[fd.GetName()]; !ok {
-			fds[fd.GetName()] = fd
-			filenames = append(filenames, fd.GetName())
-		}
+	filenames, fds, err := getFileDescriptors(symbols, descSource)
+	if err != nil {
+		return err
 	}
 	// now expand that to include transitive dependencies in topologically sorted
 	// order (such that file always appears after its dependencies)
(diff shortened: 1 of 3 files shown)


desc_source.go
```diff
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/jhump/protoreflect/desc/protoprint"
```
dragonsinth, line 7: can you move this down into the next import block?

dragonsinth (commented):
Seems reasonable to me, @jhump wdyt?

desc_source.go
Eitol: solved

desc_source.go
```diff
@@ -302,3 +294,75 @@ func addFilesToSet(allFiles []*descriptorpb.FileDescriptorProto, expanded map[st
...
+		if err := os.MkdirAll(outFilepath, 0755); err != nil {
+			return fmt.Errorf("failed to create directory %q: %v", outFilepath, err)
+		}
+		fileName := filepath.Base(fdFQName)
+		filePath := filepath.Join(outFilepath, fileName)
+		f, err := os.Create(filePath)
+		if err != nil {
+			if f != nil {
+				_ = f.Close()
+			}
```
dragonsinth, line 327: you should be able to omit this block, if err is non-nil, f will be nil

desc_source.go
```diff
@@ -302,3 +294,75 @@ func addFilesToSet(allFiles []*descriptorpb.FileDescriptorProto, expanded map[st
...
+			return fmt.Errorf("failed to create file %q: %v", filePath, err)
+		}
```
dragonsinth, line 329: immediately after this block, you should probably `defer f.Close()`

desc_source.go
```diff
@@ -302,3 +294,75 @@ func addFilesToSet(allFiles []*descriptorpb.FileDescriptorProto, expanded map[st
...
+		if err := pr.PrintProtoFile(fd, f); err != nil {
+			_ = f.Close()
+			return fmt.Errorf("failed to write file %q: %v", filePath, err)
+		}
+		_ = f.Close()
```
dragonsinth, line 334: You actually want to error check the close as well.  Something like:
```go
err = pr.PrintProtoFile(fd, f)
if err == nil {
  err = f.Close()
}
// do the error check now
```
you also don't need the final close if you deferred close earlier

desc_source.go
Eitol: solved

desc_source.go
Eitol: ready

desc_source.go
Eitol: you can't defer in a for loop.

See number 2: https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01

dragonsinth: You have to scope it sometimes:
```go
for _, fd := range allFilesSlice {
	fdFQName := fd.GetFullyQualifiedName()
	dirPath := filepath.Dir(fdFQName)
	outFilepath := filepath.Join(outProtoDirPath, dirPath)
	if err := os.MkdirAll(outFilepath, 0755); err != nil {
		return fmt.Errorf("failed to create directory %q: %v", outFilepath, err)
	}
	fileName := filepath.Base(fdFQName)
	filePath := filepath.Join(outFilepath, fileName)
	err := func() error {
		f, err := os.Create(filePath)
		if err != nil {
			return fmt.Errorf("failed to create")
		}
		defer f.Close()
		if err := pr.PrintProtoFile(fd, f); err != nil {
			return fmt.Errorf("failed to write")
		}
		if err := f.Close(); err != nil {
			return fmt.Errorf("failed to close")
		}
		return nil
	}()
	if err != nil {
		return fmt.Errorf("file %q: %w", filePath, err)
	}
}
return nil
```

Eitol:
@dragonsinth hello. Is there anything else to complete?

desc_source.go
dragonsinth: ^^ yeah this

Eitol: ready

desc_source.go
```diff
@@ -302,3 +294,79 @@ func addFilesToSet(allFiles []*descriptorpb.FileDescriptorProto, expanded map[st
...
+	f, err := os.Create(filePath)
+	if err != nil {
+		return fmt.Errorf("failed to create proto file: %v", err)
+	}
+	defer f.Close()
+	if err := pr.PrintProtoFile(fd, f); err != nil {
+		return fmt.Errorf("failed to write proto file: %v", err)
+	}
+	return nil
+}
```
dragonsinth, line 341: Let's do one more pass on this, I have a few changes:
- simplify filepath calculations and naming
- improve errors with filename and error wrapping
- 0777 so the system level umask can take effect

```go
func writeProtoFile(outProtoDirPath string, fd *desc.FileDescriptor, pr *protoprint.Printer) error {
	outFile := filepath.Join(outProtoDirPath, fd.GetFullyQualifiedName())
	outDir := filepath.Dir(outFile)
	if err := os.MkdirAll(outDir, 0777); err != nil {
		return fmt.Errorf("failed to create directory %q: %w", outDir, err)
	}

	f, err := os.Create(outFile)
	if err != nil {
		return fmt.Errorf("failed to create proto file %q: %w", outFile, err)
	}
	defer f.Close()
	if err := pr.PrintProtoFile(fd, f); err != nil {
		return fmt.Errorf("failed to write proto file %q: %w", outFile, err)
	}
	return nil
}
```

Eitol: ready

dragonsinth (approved):

jhump:
> Seems reasonable to me, @jhump wdyt?

Sorry for lack of reply. I've been on a family vacation. FWIW, looks good to me, too 👍.

rrfheff:
keen to use this feature too, when will we release it?

victorphamdeveloper:
Keen to use this feature too. Wonder how people deal with this without this feature

jhump:
@rrfheffm, @victorphamdeveloper, looks like this was included in the 1.9.2 release the other week.

PR #481, opened 2024-09-04
feat: 166: Add proxy dialer support.

nmische:
This PR ports the internal `grpc-go` dialer with proxy support into the `grpcurl` package such that `grpcurl` can support idiomatic proxy configuration via the `HTTPS_PROXY` environment variable. The port does have a minor modification in that it uses the standard lib's `net.Dialer` rather than the `NetDialerWithTCPKeepalive` from the `grpc-go`'s internal package.

This is an alternative to PR #480.

Addresses #166.

### DIFF
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -37,6 +37,8 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+var GrpcurlUA string
+
 // ListServices uses the given descriptor source to return a sorted list of fully-qualified
 // service names.
 func ListServices(source DescriptorSource) ([]string, error) {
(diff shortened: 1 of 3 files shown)


nmische:
NOTE: I not ported any of the tests over, but I can do that if this approach is preferable to #480.

proxy.go
```diff
@@ -0,0 +1,156 @@
...
+	proxyURL, err := mapAddress(addr)
+	if err != nil {
+		return nil, err
+	}
+	if proxyURL != nil {
+		newAddr = proxyURL.Host
+	}
+
+	// NOTE: Use net.Dialer to avoid dependency on grpc-go's internal package
+	// conn, err := internal.NetDialerWithTCPKeepalive().DialContext(ctx, "tcp", newAddr)
```
dragonsinth, line 137: do we need to then manually set keepalive?

grpcurl.go
```diff
@@ -37,6 +37,8 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+var GrpcurlUA string
+
```
dragonsinth, line 41: this seems.... unfortunate...

proxy.go
nmische: I don't think so. The current `grpcurl` dialer doesn't set keepalive and the `grpc-go` dailer only sets keepalive for select platforms, others get a [vanilla dialer](https://github.com/grpc/grpc-go/blob/master/internal/tcp_keepalive_others.go#L26).

PR #487, opened 2024-09-27
feat 166: support http and https proxy from environment variable of http_proxy and https_proxy

LinPr:
## description
support  http and https proxy, the added proxy file literally copyed form grpc-go source file 
[https://github.com/grpc/grpc-go/blob/master/internal/transport/proxy.go](https://github.com/grpc/grpc-go/blob/master/internal/transport/proxy.go),
 and just modify the dialer, if env http_proxy or https proxy is defined, it will try to dial the proxy address rather than the real target address
 
## relate
#480 
https://github.com/fullstorydev/grpcurl/issues/166

### DIFF
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -653,6 +653,13 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 		// handshake). And that would mean that the library would send the
 		// wrong ":scheme" metaheader to servers: it would send "http" instead
 		// of "https" because it is unaware that TLS is actually in use.
+		if network == "tcp" {
+			conn, err := proxyDial(ctx, address)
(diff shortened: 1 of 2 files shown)


nmische:
This is very similar my proposal in #481. That said I feel like #480 I may be a better approach.

LinPr:
> This is very similar my proposal in #481. That said I feel like #480 may be the better approach.
Ohh yeah，anyway，not matter what kind of solution is, the proxy is really a essential function that the grpcurl tool needs to implement. It can be applied to many scenarios!

dragonsinth:
https://github.com/fullstorydev/grpcurl/pull/480

Issue #496, opened 2024-12-09
Unix socket support broken in 1.9.2


mlafeldt:
Hey, I've been using grpcurl for a while now. A few days ago, I noticed that I can no longer connect to my gRPC server over Unix domain sockets:

```
❯ grpcurl -unix -plaintext -authority dummy /tmp/actions.sock list
Failed to dial target host "/tmp/actions.sock": connection error: desc = "transport: error while dialing: dial tcp: address /tmp/actions.sock: missing port in address"

❯ grpcurl -version
grpcurl 1.9.2
```

Version 1.9.1 works fine, so this problem must have been introduced with the latest release:

```
❯ ~/Downloads/grpcurl_1.9.1_osx_arm64/grpcurl -unix -plaintext -authority dummy /tmp/actions.sock list
grpc.reflection.v1.ServerReflection
...
```

mlafeldt:
Looking at the changelog, #480 might be the culprit, but I'm only speculating.

stuartcarnie:
Can also confirm that 1.9.1 works and 1.9.2 is broken (macOS)

stuartcarnie:
I can confirm that reverting #480 resolves the issue

zhyuri:
I think the change #480 is correct. In fact, with that commit, the `network` parameter can be removed from `BlockingDial` along with the `-unix` option from the CLI.

This is because the grpc-go dialer can handle RFC 3986 compliant URI. In the example from @mlafeldt, I believe the following should work in grpcurl 1.9.2. Note that `-unix` is removed and the proxy address has a schema `unix://`
```bash
grpcurl -plaintext -authority dummy unix:///tmp/actions.sock list
```

jhump:
@zhyuri, the changes you propose (and that are in PR #498) are not backwards compatible, so I think are non-starter. It would be better to re-formulate this so it's a backwards-compatible change, in a way that also fixes the `-unix` flag. I left comments on the PR with more details.

zhyuri:
@jhump thanks for the review!

I pushed a few commits to the PR to address your comments. I also updated the PR description to reflect the change.

PR #498, opened 2024-12-19
Restore Unix socket support

zhyuri:
Fixes https://github.com/fullstorydev/grpcurl/issues/496

## Context

https://github.com/fullstorydev/grpcurl/pull/480 introduced a behaviour change by replacing custom dialer with the default one from grpc-go, where `grpcurl -unix /path/to/proxy` no longer works because the address isn't a RFC 3986 compliant URI for unix sockets.
The fix aims to restore the previous behaviour in both grpcurl CLI and package API `BlockingDial`.

### CLI

* Prepend `unix://` to the address if `-unix` option is used but the address is not a RFC 3986 compliant URI. This change restored the previous behaviour for `-unix`.
* Removed default authority `localhost` for `-unix`. This default option was added in https://github.com/fullstorydev/grpcurl/commit/149a93e0ec14b6fa2b80b6c47c9348186f343af4 in order to align the behaviour with popular grpc clients/servers in the customized dialer. Because we now use the grpc-go default dialer, which [sets](https://github.com/grpc/grpc-go/blob/56a14ba1f818a0ca297f5e21bf689064a714ee68/clientconn.go#L1798-L1799) `localhost` as the [default authority for unix networktype](https://github.com/grpc/grpc-go/blob/56a14ba1f818a0ca297f5e21bf689064a714ee68/internal/resolver/unix/unix.go#L64-L66), this default option is no longer needed.
*  Always use empty string as `network` in `grpcurl.BlockingDial`.

### grpcurl package

Depending on the `network`, the `BlockingDial` API performs different actions
* When empty, use address as is. CLI prepends `unix://` to the `-unix` address and always sets network to empty.
* When `tcp`, validate the address doesn't have `unix://` prefix.
* When 'unix`, prepend `unix://` if missing.
* Otherwise use custom dialer with provided network.

## Test

- [x] `grpcurl -unix unix:///path/to/proxy` - works
- [x] `grpcurl -unix /path/to/proxy` - works, this is the behaviour before #480 
- [x] `grpcurl unix://path/to/proxy` - works
- [x] `grpcurl /path/to/proxy` - does not work with expected error `missing port in address`


cmd/grpcurl/grpcurl.go
```diff
@@ -363,6 +363,9 @@ func main() {
 	if *format != "json" && *format != "text" {
 		fail(nil, "The -format option must be 'json' or 'text'.")
 	}
+	if isUnixSocket != nil && isUnixSocket() {
+		fail(nil, "The -unix option is deprecated, please use unix://{path} as the address instead.")
+	}
```
jhump, line 368: I don't think it should fail the command. That would be backwards-incompatible. Instead, it should add the "unix://" prefix to the target addres (if it's not already present). I don't actually think we even need to deprecate the flag -- we could just keep it as a synonym of sorts for using a "unix://" prefix. (See other comment for more.)

grpcurl.go
```diff
@@ -609,7 +609,7 @@ func ServerTransportCredentials(cacertFile, serverCertFile, serverKeyFile string
 // BlockingDial is a helper method to dial the given address, using optional TLS credentials,
 // and blocking until the returned connection is ready. If the given credentials are nil, the
 // connection will be insecure (plain-text).
-func BlockingDial(ctx context.Context, network, address string, creds credentials.TransportCredentials, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
+func BlockingDial(ctx context.Context, address string, creds credentials.TransportCredentials, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
```
jhump, line 612: This is a backwards-incompatible change. This is exported API and this module is post-v1.0. Changing the signature would require a v2.0, which is way more effort than is warranted for such a small change.

I think the following behavior would do what we need:
1. If network is the empty string, just use address as is.
2. If network is "tcp", return an error if address has a "unix://" prefix.
3. If network is "unix" and address does not have a "unix://" prefix, add one.
4. Otherwise, configure a custom dialer to actually dial using the selected network name.

So the command-line `main` package would always pass the empty string. But this way, we haven't broken any existing callers, including those that may be using other values (such as "tcp4" or "tcp6" to disable dual-stack support).

cmd/grpcurl/grpcurl.go
```diff
@@ -486,6 +483,13 @@ func main() {
 		if *maxMsgSz > 0 {
 			opts = append(opts, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(*maxMsgSz)))
 		}
+		if isUnixSocket != nil && isUnixSocket() && !strings.HasPrefix(target, "unix://") {
+			// prepend unix:// to the address if it's not already there
+			// this is to maintain backwards compatibility because the custom dialer is replaced by
+			// the default dialer in grpc-go.
+			// https://github.com/fullstorydev/grpcurl/pull/480
+			target = "unix://" + target
+		}
```
jhump, line 492: You could instead set a `network` local variable (that defaults to empty string) to `"unix"` and then pass to `BlockingDial` below. That way, you aren't duplicating the logic that also lives in `BlockingDial` for this scenario.

jhump (commented):
Looks pretty good to me. I'm not actually a maintainer anymore, so you'll also need a review from the likes of @dragonsinth or @gpassini.

PR #502, opened 2025-01-10
Update to v1.17.0 of jhump/protoreflect

### LINKED ISSUE #478: Version 1.9.1: Failed to process proto source files
I was generating REST API endpoint, using gRPC Gateway, by adding [custom annotations](https://github.com/grpc-ecosystem/grpc-gateway?tab=readme-ov-file#2-with-custom-annotations), which requeired including some vendor `.proto` files, such as [descriptor.proto](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto). After adding annotations, I successfully generated `.pb.gw.go` files and calls to REST API worked just fine.

But calls to gRPC API, using `grpcurl`, stopped working:
```
Failed to process proto source files.: could not parse given files: google/protobuf/descriptor.proto:1070:5: extension range cannot have declarations and have verification of UNVERIFIED
```

Then I tried other versions of grpcurl and figured out that same calls wokred with versions from `1.8.2` to `1.8.9`.

jhump:
This should fix #478.


dragonsinth:
Oh no, this fails our static checks I guess because some stuff got deprecated and is directly referenced from grpcurl?

jhump:
Ah, yeah. I think I just need to add some nolint annotations or the like... (I deprecated all of the stuff that is now redundant with stuff in `google.golang.org/protobuf`, hoping that v1.17.0 would be a "final" v1 release...)

dragonsinth (approved):

dragonsinth:
Sweet, thanks!

PR #512, opened 2025-03-13
snap

PietroPasotti:
Hi there! 

I'd like to release a snap for grpcurl, so we can more easily use it in our CI and testing environments.
Would you be happy with me contributing this to the repo?
If you're not interested in this, I could always register the snap under my own name, but that might be confusing for users who see an unexpected publisher for it.

Fixes #199

PR #514, opened 2025-03-14
Snap: strict confinement, stable grade and docs

PietroPasotti:
Hi!
On top of #512, I am hereby:
- locking down the confinement of the snap to `classic`, this means that the installer has [certain strong guarantees](https://snapcraft.io/docs/snap-confinement) that the package won't do undesired things to their system
- upgrading the grade to `stable`, so that the snap can be released and installed by anyone
- adding a badge to the readme and a note about installing the snap
- a readme so any dev working on this project knows what the snapcraft.yaml is about and how to maintain it

I have released a snap for the current 1.9.3 tag

image

This means that if you run `sudo snap install grpcurl --channel edge`, you'll have grpcurl 1.9.3 within seconds up and running in your machine!
Which IMHO is great.

The next step, if you're interested, would be to set up some automation so that every time you create a new release, we also release a new snap to the edge channel (and what previously was on edge can be promoted up towards a stable channel).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ dist/
 .idea/
 VERSION
 .tmp/
+*.snap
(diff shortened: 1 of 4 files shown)


dragonsinth (approved):
sure why not 😄

PR #540, opened 2025-12-23
docs(README): add mise alternate installation documentation

jylenhof:


dragonsinth:
I don't think we need to embed an ad for mise here-- existing mise users should be able to find tools they want through mise.

PR #542, opened 2026-01-22
Add Windows Arm64 release

### LINKED ISSUE #541: Create releases for Windows Arm64
Hi, thank you for creating and maintaining this tool! We plan to use it in an application to dynamically call gRPC services.

I noticed that currently there is no release for **Windows Arm64**, whereas there are Arm64 releases for other OSes such as Linux. Can this be added?
(While amd64 executables can run on Windows Arm64 using emulation, running native Arm64 executables should have better performance.)

I don't know much about Go, but according to the [release notes of **Go 1.17**](https://tip.golang.org/doc/go1.17#windows), Windows Arm64 should be supported starting with that version.

Thanks!

kpreisser:
Try to add Windows Arm64 releases by removing the `ignore` entry for that combination.

Windows Arm64 should be supported starting with [Go 1.17](https://tip.golang.org/doc/go1.17#windows).

Fixes #541

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -19,8 +19,6 @@ builds:
     ignore:
       - goos: darwin
         goarch: 386
-      - goos: windows
-        goarch: arm64
       - goos: darwin
         goarch: arm
       - goos: windows


dragonsinth (commented):
@gpassini I think you've done the most recent releases... any issues with this?

gpassini (approved):
LGTM!

Issue #387, opened 2023-04-20
grpcurl fails with "context deadline exceeded" after 10s if using plaintext when server expects TLS


ucarion:
Bottom line up front, here's how you reproduce this issue:

```
$ grpcurl -version
grpcurl 1.8.6

$ time grpcurl -plaintext grpcb.in:9001 list
Failed to dial target host "grpcb.in:9001": context deadline exceeded
grpcurl -plaintext grpcb.in:9001 list  0.02s user 0.03s system 0% cpu 10.082 total
```

For context, `grpcb.in:9001` wants TLS; `-plaintext` is the problem. But the fact that grpcurl hangs for 10 seconds, and does not produce an informative error, is the subject of this GitHub issue. I suspect the issue may be use of `grpc.WithBlock()` prevents an error from bubbling up, but I assume there's a good reason for the use of that dialopt for some other purpose.

jhump:
>  I suspect the issue may be use of `grpc.WithBlock()` prevents an error from bubbling up,

I doubt that. That's actually _how_ you get any error to bubble up. Otherwise, you never get any sort of feedback from `Dial` as it does the actual TCP connection setup completely asynchronous and only returns an error if there is some other configuration problem with the options.

The issue here is _where_ it fails. In `grpcurl.BlockingDial`, we try to control both dialing and a potential TLS handshake so that we can intercept any errors (which the underlying gRPC Go runtime library hides from the application), in order to give a decent error to the user.

The issue here is actually that the connections are setup just fine -- all a plaintext connection cares about is getting the TCP connection. The other direction (using TLS in the client to a server that does not expect it) fails more cleanly because the error _does_ bubble up from dialing because the connections cannot be established because the TLS handshake fails.

So the actual error is happening inside the gRPC runtime when it tries to send the [HTTP/2 preface to the server](https://httpwg.org/specs/rfc7540.html#ConnectionHeader). In this case, the server is expecting a TLS handshake, but doesn't receive one. So the server immediately closes the connection. We're providing a `grpc. FailOnNonTempDialError(true)` dial option, in the hopes that something like this would be bubbled up from the dial call. But apparently the server suddenly closing the connection (without any known reason) is interpreted as a temporary error. So the runtime keeps re-trying, creating a new connection over and over, never getting a healthy one that can be used for sending an RPC.

[...]

PR #564, opened 2026-06-06
Add support for TLS 1.3

### LINKED ISSUE #563: TLS 1.3 is not supported
The maximum supported TLS version is set to 1.2, in:

https://github.com/fullstorydev/grpcurl/blob/cb68aaa7f5b2b9ea5bf6ab96d0f34e100e946147/grpcurl.go#L571-L573

The referenced issue https://github.com/golang/go/issues/28779 is closed as the behavior of crypto/tls is as expected.

I cannot see the linked build (I assume Travis cleaned it up) but I assume the undesired behavior is that when the current code (when allowing TLS1.3) sends a plaintext request to a server that requires a client cert, that the server rejects it without the client noticing, so the client just retries/hangs. I assume "fixing" just means letting the client return a clear error to the user quickly.

I would like to use TLS 1.3, and I'm happy to send a PR.

bcleenders:
This PR allows TLS 1.3, by removing the MaxVersion in the client config.

This would silently swallow errors, so e.g. a client without cert dialing a server that requires client certs would lead to an error which gets ignored, leading to retries until timeout.

In this PR, we wrap the connection and if an error occurs we send it to the existing `result` channel.

I think this matches @jhump's comment in https://github.com/fullstorydev/grpcurl/issues/387#issuecomment-1517098394 (but for this PR, I scoped the change to TLS only to avoid dropping a huge patch)

 ### Testing

Start a test server that requires client certs
```console
# Start the test server (in another tab)
$ go run ./internal/testing/cmd/testserver \
    -cert internal/testing/tls/server.crt \
    -key internal/testing/tls/server.key \
    -cacert internal/testing/tls/ca.crt \
    -requirecert -p 9999
```

Old behavior
```console
$ grpcurl -cacert internal/testing/tls/ca.crt localhost:9999 list
Failed to dial target host "localhost:9999": context deadline exceeded
```

New behavior
```console
$ go run ./cmd/grpcurl -cacert internal/testing/tls/ca.crt localhost:9999 list
Failed to dial target host "localhost:9999": remote error: tls: certificate required
exit status 1
```

The old behavior is to hang until we hit the deadline. The new behavior is to return immediately with an error.

Fixes #563

### DIFF
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -20,6 +20,7 @@ import (
 	"slices"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/golang/protobuf/proto"   //lint:ignore SA1019 we have to import these because some of their types appear in exported API
 	"github.com/jhump/protoreflect/desc" //lint:ignore SA1019 same as above
@@ -568,9 +569,6 @@ func ClientTLSConfig(insecureSkipVerify bool, cacertFile, clientCertFile, client
 // client certs. The serverCertFile and serverKeyFile must both not be blank.
 func ServerTransportCredentials(cacertFile, serverCertFile, serverKeyFile string, requireClientCerts bool) (credentials.TransportCredentials, error) {
 	var tlsConf tls.Config
-	// TODO(jh): Remove this line once https://github.com/golang/go/issues/28779 is fixed
-	// in Go tip. Until then, the recently merged TLS 1.3 support breaks the TLS tests.
-	tlsConf.MaxVersion = tls.VersionTLS12
 
 	// Load the server certificates from disk
 	certificate, err := tls.LoadX509KeyPair(serverCertFile, serverKeyFile)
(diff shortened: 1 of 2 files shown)


grpcurl.go
```diff
@@ -734,6 +732,34 @@ func (c *errSignalingCreds) ClientHandshake(ctx context.Context, addr string, ra
...
+	return conn, auth, nil
+}
+
+// errSignalingConn wraps a net.Conn to capture the first read error and
+// report it via writeResult. This allows BlockingDial to surface post-handshake
+// errors.
+type errSignalingConn struct {
+	net.Conn
+	writeResult func(res interface{})
+	once        sync.Once
```
dragonsinth: I don't think we need the sync.Once, writeResult() has to be idempotent. If we're worried about spamming that `select` statement too many times, we should add an atomic CompareAndSwap guard inside writeResult(), and then unconditionally push and close the channel

bcleenders:
Thanks for the review! I've implemented your suggestion.

I'm also happy to apply the same connection-error handling logic to apply to plaintext connections, if you'd like (by getting rid of [this if statement](https://github.com/fullstorydev/grpcurl/pull/564/changes#diff-f26eee6adb3266e2661632d07bd47a3dbf12d9186fd118a199623cc86cbfe686R741-R743)). If that sounds good, I can do it as a follow-up PR, as some tests need updating if I make that change.

dragonsinth (approved):

dragonsinth:
I think it's fine to just support TLS here

dragonsinth:
But feel free to follow up if you can make better error reporting on plain text happen!

dragonsinth:
FYI: we've been getting test flakes recently--

```
--- FAIL: TestBrokenTLS_RequireClientCertButNonePresented (0.02s)
    tls_settings_test.go:332: expecting a TLS certificate error, got: read tcp 127.0.0.1:49228->127.0.0.1:42915: use of closed network connection
```
It's either this PR, or #567 - unsure. Would you mind taking a look @bcleenders ?

bcleenders:
hi! Sorry, I hadn't noticed your message here. Yes, looking into it!

PR #566, opened 2026-06-08
fix xds credentials being silently ignored

### LINKED ISSUE #565: xDS credentials silently ignored
#424 implemented xDS credentials for channel targets that start with xds:/// but when testing it doesn't work since grpc-go has this check:

https://github.com/grpc/grpc-go/blob/caf0772c2bcb8bc15d43eb53448e921f34f0b7e8/internal/xds/balancer/clusterimpl/clusterimpl.go#L97-L99 

and `errSignalingCreds` in grpcurl.go doesn't implement UsesXDS().

Also, a channel target like `xds://authority.example.com/service-foobar` is valid so the prefix check should be "xds://" (two slashes).

I'll open a PR.

erikjoh:
Fixes #565

Before change:
```
GRPC_XDS_BOOTSTRAP=/tmp/b.json go run cmd/grpcurl/grpcurl.go xds:///<target service> list
Failed to dial target host "xds:///<target service>": tls: failed to verify certificate: x509: certificate is not valid for any names, but wanted to match <target service>
exit status 1
```

After change:
```
GRPC_XDS_BOOTSTRAP=/tmp/b.json go run cmd/grpcurl/grpcurl.go xds:///<target service> list
grpc.health.v1.Health
grpc.reflection.v1.ServerReflection
grpc.reflection.v1alpha.ServerReflection
...
```

### DIFF
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -618,7 +618,7 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 
 	var err error
 	if strings.HasPrefix(address, "xds://") {
-		// The xds:/// prefix is used to signal to the gRPC client to use an xDS server to resolve the
+		// The xds:// prefix is used to signal to the gRPC client to use an xDS server to resolve the
 		// target. The relevant credentials will be automatically pulled from the GRPC_XDS_BOOTSTRAP or
 		// GRPC_XDS_BOOTSTRAP_CONFIG env vars.
 		creds, err = xdsCredentials.NewClientCredentials(xdsCredentials.ClientOptions{FallbackCreds: creds})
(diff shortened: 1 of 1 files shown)


erikjoh:
@bcleenders I assume you also would be interested in this :grin:

grpcurl.go
```diff
@@ -617,7 +617,7 @@ func BlockingDial(ctx context.Context, network, address string, creds credential
 	}
 
 	var err error
-	if strings.HasPrefix(address, "xds:///") {
+	if strings.HasPrefix(address, "xds://") {
 		// The xds:/// prefix is used to signal to the gRPC client to use an xDS server to resolve the
```
dragonsinth:
```suggestion
		// The xds:// prefix is used to signal to the gRPC client to use an xDS server to resolve the
```

erikjoh: Fixed!

dragonsinth (approved):

PR #567, opened 2026-06-12
Fast-fail on connection errors

### LINKED ISSUE #387: grpcurl fails with "context deadline exceeded" after 10s if using plaintext when server expects TLS

bcleenders:
Avoid hanging on connection errors; fail fast by propagating connection errors (not just for TLS).

This also speeds up tests a lot (from >20s to about 3s), since the tests included some connection timeouts.

This expands on https://github.com/fullstorydev/grpcurl/pull/564 ; it does the same for plaintext connections.

Fixes #387

### Demo

Start server that expects TLS (to mimick the issue linked above):

```
$ go run ./internal/testing/cmd/testserver \
    -cert internal/testing/tls/server.crt \
    -key internal/testing/tls/server.key \
    -p 50051
```

Without fix:
```console
$ grpcurl -plaintext localhost:50051 list
Failed to dial target host "localhost:50051": context deadline exceeded
```

With fix:
```console
$ ./grpcurl -plaintext localhost:50051 list
Failed to dial target host "localhost:50051": read tcp 127.0.0.1:64984->127.0.0.1:50051: read: connection reset by peer
```

### DIFF
--- a/grpcurl.go
+++ b/grpcurl.go
@@ -733,20 +733,14 @@ func (c *errSignalingCreds) ClientHandshake(ctx context.Context, addr string, ra
 		c.writeResult(err)
 		return conn, auth, err
 	}
-	// Wrap TLS connections to capture post-handshake errors. With TLS 1.3,
-	// client certificate rejection by the server happens after the client
-	// considers the handshake complete. The server's TLS alert surfaces on the
-	// first Read from the connection. Only TLS connections need this (plaintext
-	// connections don't have post-handshake alerts).
-	if _, isTLS := auth.(credentials.TLSInfo); isTLS {
-		conn = &errSignalingConn{Conn: conn, writeResult: c.writeResult}
-	}
-	return conn, auth, nil
+	// Wrap the connection to capture post-handshake errors, e.g.:
+	// - TLS 1.3 client cert rejection (server sends alert after handshake)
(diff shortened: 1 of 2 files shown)


dragonsinth (approved):
nice improvement!

dragonsinth:
FYI: we've been getting test flakes recently--

```
--- FAIL: TestBrokenTLS_RequireClientCertButNonePresented (0.02s)
    tls_settings_test.go:332: expecting a TLS certificate error, got: read tcp 127.0.0.1:49228->127.0.0.1:42915: use of closed network connection
```
It's either this PR, or #564 - unsure. Would you mind taking a look @bcleenders ?

PR #573, opened 2026-07-27
Fix path traversal attack in `-proto-out-dir`

bcleenders:
A malicious file descriptor name (e.g. from an untrusted protoset or server reflection) could escape the output directory through `../` sequences, writing attacker-controlled content to arbitrary paths.

Confine all writes to an `os.Root` opened on the output directory.

---

> [!NOTE]
> **Low Risk**
> Security-hardening change with targeted tests; behavior for legitimate relative proto paths should be unchanged while blocking traversal.
> 
> **Overview**
> **Fixes a path traversal vulnerability** when exporting protos via `-proto-out-dir`: descriptor names from untrusted protosets or server reflection could use `../` and write outside the chosen output directory.
> 
> Proto export now opens an **`os.Root`** on the output directory and performs all `MkdirAll` / `Create` operations through it, so malicious names like `../escape.proto` or nested `foo/../../../escape.proto` are rejected instead of escaping the sandbox. The write loop is factored into **`writeProtoFiles`**, and directory creation uses **`0755`** instead of **`0777`**.
> 
> **Tests** cover a normal nested path (`foo/bar.proto`) and assert traversal attempts fail and do not create files outside the temp output dir.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4fbb35359081edadbff214061bdb5c72b49ba07. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dragonsinth (approved):
LGTM after a few changes; I wanted to actually traverse the live-code root

